PR #2, opened 2024-08-06
feat: implement HoneycombOptions

beekhc:

Implements `HoneycombOptions`, including `HoneycombOptions.Builder`, as per the design doc.

* If a plist file is passed into the constructor, `Builder` is pre-populated with its values.
* There are setters for all values.
* `HoneycombOptions` only contains specific values like `tracesEndpoint`, whereas the `Builder` has fallback settings like `setEndpoint`. The specific values are resolved when `build()` is called.
* Automatic headers and resources are injected upon calling `build()`.
* The only required field is `APIKey`.
* Throws HoneycombOptionsError on any fatal configuration error. These should only be unexpected developer bugs, not expected runtime errors.

There are comprehensive unit tests that I believe test all plist values, setters, and fallbacks, as well as any error conditions.

In review please look for the following:
* Are there any missing unit test cases?
* Is the fallback logic correct?
* Are the resources and headers auto-populated correctly?
* Should the error cases actually be error cases?


pkanal (commented):
This looks great! I think the only thing we're missing here is that _if_ the user using a classic API key, we have to set the `x-honeycomb-dataset` header to set the `service.name`. Here's an example: https://github.com/honeycombio/honeycomb-opentelemetry-web/blob/a00769688c0402323936813096516be810401014/packages/honeycomb-opentelemetry-web/src/http-json-trace-exporter.ts#L21

Here's the part of the spec for supporting Honeycomb's classic environment: https://github.com/honeycombio/specs/blob/main/specs/otel-sdk-distro-classic.md#spanexporter

pkanal (approved):
This looks great! 🎉 

Logs and traces are the same, and metrics have the different behaviour that you've already accounted for!

PR #4, opened 2024-08-21
feat(swift): implement Honeycomb.configure and add smoke tests

beekhc:

* Adds the `Honeycomb.configure` method, which configures OpenTelemetry for Swift using `HoneycombOptions`.
* Also adds smoke tests.

The smoke-test is based on the one in honeycomb-opentelemetry-android, as we have many of the same constraints. The iOS simulator must be run on macOS, and cannot be run in docker. However, we should still be able to use CircleCI, with a machine image.

Follow setup instructions in DEVELOPING.md and then run:

```sh
make smoke
make unsmoke
```

---
  - https://app.asana.com/0/0/1207992777792927


pkanal (approved):
🎉

JamieDanielson (commented):
When I try running the `run-ios-tests.sh` script I am getting errors like 

2024-08-26 17:15:15.204607-0400 xctest[44054:15043162] [Default] Failed to create a bundle instance representing '/Users/jamiedanielson/Library/Developer/Xcode/DerivedData/honeycomb-opentelemetry-swift-gobafwzxbhoxhbeeegnvjlpxsvky/Build/Products/Debug/HoneycombTests.xctest'. Check that the bundle exists on disk.

I can't figure out what I am missing. I've tried adding a `clean` option to the `xcodebuild` command in `run-ios-tests.sh`, and also tried adding a workspace with the path `.../.swiftpm/xcode/package.xcworkspace` but those don't seem to work.

If I list the workspace I can see resolved source packages that make sense to me, so I think most of it is set up properly. I'll keep looking to see what else it could be but open to hearing any ideas you might have!

JamieDanielson (approved):
A few suggestions but they are mostly nits. Took me a while to get this running because I am unfamiliar with the tooling, but once I got there it all works 😄

PR #5, opened 2024-08-30
feat(swift): add MetricKit support

beekhc:

This adds MetricKit signals to the Honeycomb Swift SDK.

This adds all of the data from Apple's MetricKit into Honeycomb. It's not obvious how to structure this data in order to best fit OTel semantics conventions while also being most usable in Honeycomb. I made my best guess how to do it, but would appreciate feedback on it. I am really not sure if this is the best way to represent this data. Here is how I'm structuring the data:

### MetricKit

MetricKit reports a batch of data approximately once a day, with cumulative data from the previous day. There are two big categories of data: _Metrics_ and _Diagnostics_. These are covered below.

All of this data is captured using the scope `"@honeycombio/instrumentation-metric-kit"`

### Units

MetricKit measurements have units, so that a metric might report "1 kb" or "3 hours". To represent this data in attributes, they are all converted to the standard unit for their type (such as bytes or seconds) and represented as doubles.

### Metrics

Metrics are things like "total cpu time" and "number of abnormal exists". They are pre-aggregated over the reported period. For metrics, we report a single `Span` named `"MXMetricPayload"`. The start time of the span is the start of the reporting period, and the end of the span is the end time of the reporting period. So these spans are generally 24 hours long. Each metric is included as an attribute on that span. The attributes are namespaced corresponding to their category in `MXMetricPayload`. For example, the metrics mentioned above are named `metrickit.app_exit.foreground.abnormal_exit_count` and `metrickit.cpu.cpu_time`. For histogram data, we include only the estimated average value for now.

I used an OTel span instead of OTel metrics for a few reasons:
* Metrics usually go into a different dataset, and we want all the data in one place for a launchpad.
* The data is pre-aggregated, not individual values, so some of the OTel metric semantics don't make sense.
* The time the data is reported doesn't accurately reflect when the events occur. For example, we don't want to interpret 1000 cpu instructions over 24 hours as sitting idle for 23.999 hours and then suddenly bursting 1000 instructions.
* The Swift metrics APIs are confusing, and it's not clear how to represent some of the data.

We could add OTel metrics later too, if we want to.

### Diagnostics

Diagnostics are events that occur once. However, we don't know exactly when. We only know they occurred during the reporting period. These include things like crashes and hangs. To represent diagnostics, we emit a top-level `Span` called `"MXDiagnosticPayload"` with the start and end corresponding to the reporting period. Then, for each event, we emit an OTel "log" for the event. I used logs instead of traces, because each event is instantaneous as far as we know. Also, I wanted to be consistent with the way the Android OTel auto-instrumentation represents uncaught exceptions. Each log has a name attribute for its type, with the key `"metrickit.diagnostic.name"`. The logs have metadata attached with namespaced keys, similar to metrics. For example, crash exception codes have the key `"metrickit.diagnostic.crash.exception.code"`. Since we don't know the exact times of the events, all logs use the end time of the reporting period. This helps so that data shows up as new data in Honeycomb when it arrives.

There are copious smoke tests to make sure that all of the expected values are present, and that all values are using the correct units, since that's an easy mistake to make when working with the Measurement class. They probably look like overkill, but already caught several fields I had overlooked during development.

`make smoke`

Apple makes it very difficult to test MetricKit in general. It does not work in the simulator at all. With physical devices in debug mode, there is a menu option to send a simulated payload, but it always sends the same values. And a real app only sends a batch once a day. I have manually tested this with the simulated payload.

---
  - https://app.asana.com/0/0/1207968470870231

### DIFF
--- a/Sources/Honeycomb/Honeycomb.swift
+++ b/Sources/Honeycomb/Honeycomb.swift
@@ -1,6 +1,7 @@
 
 import Foundation
 import GRPC
+import MetricKit
 import NIO
 import OpenTelemetryApi
 import OpenTelemetrySdk
@@ -35,7 +36,9 @@ private func createKeyValueList(_ dict: [String: String]) -> [(String, String)] 
     return result
 }
 
-class Honeycomb {
+public class Honeycomb {
+  static private let metricKitSubscriber = MetricKitSubscriber()
+  
     static public func configure(options: HoneycombOptions) throws {
         guard let tracesEndpoint = URL(string:options.tracesEndpoint) else {
             throw HoneycombOptionsError.malformedURL(options.tracesEndpoint)
@@ -143,5 +146,7 @@ class Honeycomb {
         OpenTelemetry.registerTracerProvider(tracerProvider: tracerProvider)
         OpenTelemetry.registerMeterProvider(meterProvider: meterProvider)
         OpenTelemetry.registerLoggerProvider(loggerProvider: loggerProvider)
+
+      MXMetricManager.shared.add(self.metricKitSubscriber)
     }
 }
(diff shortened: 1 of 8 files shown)


wolfgangcodes (approved):
🚢 it!

PR #8, opened 2024-09-17
chore(swift): update for new swift format

beekhc:

Updates the Makefile and instructions for running swift format.

As of Xcode 16, released this week, `swift format` is now included automatically with Xcode. This is now the recommended way to run what was previously released as `swift-format` and had to be installed with brew.

While it would be technically possible to write a script to automatically detect `swift-format` vs `swift format`, the complexity would not be worth the hassle, as it's recommended to use the latest version of Xcode anyway.

This has the nice side effect that it should fix our circleci lint checks.

`make lint`
`make format`

---
  - https://app.asana.com/0/0/1208249202227272

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -4,7 +4,7 @@
 
 **Required:**
 
-- Xcode (from the [Mac App Store](https://apps.apple.com/us/app/xcode/id497799835))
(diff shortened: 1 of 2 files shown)


wolfgangcodes (approved):
🧹

PR #9, opened 2024-09-19
fix(macOS): make SDK build in macOS

beekhc:

This adds annotations to make the SDK work in macOS.

Supporting macOS is not a priority for us, but it's nice to have. I'm currently using it to simulate some data from a command-line app. All it takes is a few annotations, so I think it's worth doing.

Note that the diff looks a lot more complicated than the actual changes. GitHub seems to be confused about some indentation changes.

It's not worth adding a test for this at this time. But I have manually verified it by using the SDK.


wolfgangcodes (approved):

MustafaHaddara (approved):
seems reasonable enough

PR #11, opened 2024-09-23
refactor(swift): Use BatchSpanProcessor instead of SimpleSpanProcessor

pkanal:
Use `BatchSpanProcessor` instead of `SimpleSpanProcessor` since it is the recommended span processor to use in production apps.

Leaving this in draft to start because this might affect smoke tests since it might take a few seconds for the spans to batch send with this change so we might need to add in a wait mechanism.

- Use `BatchSpanProcessor` from OTel as the default span processor to send spans.

- Smoke tests still pass


beekhc (approved):
looks great!

PR #16, opened 2024-10-10
ci: Run smoke tests in CI

pkanal:
Run smoke tests in CI on every branch/PR and nightly on main.

- Add smoke_tests job to Circle CI
- Install the collector directly on the CI machine instead of through docker to run smoke tests against

- [Smoke tests run and pass on CI](https://app.circleci.com/pipelines/github/honeycombio/honeycomb-opentelemetry-swift/125/workflows/1041f6bd-8e50-4b47-9e04-84b3c2c09dea/jobs/169)
- Smoke tests still run and pass locally

Screenshot 2024-10-10 at 3 39 16 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,17 +54,15 @@ jobs:
             bats --version
       - run: xcodebuild -version
       - run:
-          name: Install Docker Compose
-          environment:
-            COMPOSE_VERSION: "v2.25.0"
+          name: Install OTel Collector
+          # environment:
+          #   COMPOSE_VERSION: "v2.25.0"
           command: |
-            curl -sSL "https://github.com/docker/compose/releases/download/${COMPOSE_VERSION}/docker-compose-$(uname -s)-$(uname -m)" -o ~/docker-compose
-            mkdir -p ~/.docker/cli-plugins
-            chmod +x ~/docker-compose
-            mv ~/docker-compose ~/.docker/cli-plugins/docker-compose
+            curl --proto '=https' --tlsv1.2 -fOL https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.111.0/otelcol_0.111.0_darwin_arm64.tar.gz
+            tar -xvf otelcol_0.111.0_darwin_arm64.tar.gz
       - run:
-          name: Build and run the collector
-          command: docker compose up --build collector --detach
+          name: Start OTel Collector
+          command: ./otelcol --config=./smoke-tests/collector/otel-collector-config.yaml
 workflows:
   nightly:
     triggers:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,6 +82,7 @@ workflows:
     jobs:
       - lint
       - tests
+      - smoke_tests
 
   build:
     jobs:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -32,6 +33,43 @@ jobs:
...
+      xcode: 15.4.0
+    resource_class: macos.m1.medium.gen1
+
+    steps:
+      - attach_workspace:
+          at: ./
+      - macos/preboot-simulator:
+          version: "17.5"
+          platform: "iOS"
+          device: "iPhone 15"
```
martin308, line 47: do we want to use some matrix tests to cover more versions/platforms/devices?

.circleci/config.yml
```diff
@@ -32,6 +33,43 @@ jobs:
...
+          command: |
+            which bats
+            bats --version
+      - run: xcodebuild -version
+      - run:
+          name: Install OTel Collector
+          environment:
+            COLLECTOR_VERSION: "0.111.0"
+          command: |
+            curl --proto '=https' --tlsv1.2 -fOL https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v${COLLECTOR_VERSION}/otelcol_${COLLECTOR_VERSION}_darwin_arm64.tar.gz
```
martin308, line 61: I was going to suggest using docker or caching the download could be good but the smoke tests take 10mins to run so probably not worth it 😂

.circleci/config.yml
beekhc: if we do test on older iOS versions, we'll have to change the test expectations, because various MetricKit signals were added in different iOS versions. it's probably not trivial

.circleci/config.yml
```diff
@@ -32,6 +33,43 @@ jobs:
...
+      - run:
+          name: Start OTel Collector
+          command: touch ./smoke-tests/collector/data.json && ./otelcol --config=./smoke-tests/collector/otel-collector-config.yaml --set="exporters::file::path=./smoke-tests/collector/data.json"
+          background: true
+      - run:
+          name: Run iOS App
+          command: bash ./run-ios-tests.sh
+      - run:
+          name: Run Smoke Tests
+          command: cd smoke-tests && bats ./smoke-e2e.bats --report-formatter junit --output ./
```
beekhc, line 72: Doing my best to channel my inner @JamieDanielson and ask, would it make sense to move any of these commands into the Makefile?

beekhc (commented):
looks good to me

.circleci/config.yml
pkanal: I think we should make this a follow up task, but overall yes this sounds like a good idea!

.circleci/config.yml
pkanal: That was the original plan 😭 but I found it was really hard to get docker working on the CI mac machine and the collector setup itself is super fast. The thing that takes the longest is running the iOS app, which we can probably make faster with some CircleCI tricks.

.circleci/config.yml
pkanal: There's already a make command for this so I will switch to that!

martin308 (approved):

beekhc (approved):

PR #17, opened 2024-10-10
feat(samplers): Add deterministic sampler

martin308:

Adds a deterministic sampler modeled off the deterministic sampler in honeycombs other distros.

unit tests

### DIFF
--- a/Sources/Honeycomb/HoneycombDeterministicSampler.swift
+++ b/Sources/Honeycomb/HoneycombDeterministicSampler.swift
@@ -1,0 +1,60 @@
+import Foundation
+import OpenTelemetryApi
+import OpenTelemetrySdk
+
+class HoneycombDeterministicSampler: Sampler {
+    private let inner: Sampler
+    private let rate: [String: AttributeValue]
+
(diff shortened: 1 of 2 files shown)


Sources/Honeycomb/HoneycombDeterministicSampler.swift
```diff
@@ -0,0 +1,60 @@
...
+    init(sampleRate: Int) {
+        var inner: Sampler
+
+        switch sampleRate {
+        case Int.min..<1:
+            inner = Samplers.alwaysOff
+        case 1:
+            inner = Samplers.alwaysOn
+        default:
+            inner = Samplers.traceIdRatio(ratio: 1.0 / Double(sampleRate))
```
pkanal, line 18: Why are we doubling the sample rathe here for the ratio? Or am I reading the code wrong 🙈

pkanal: Ah nvm I just don't know how swift works!

pkanal (approved):

PR #18, opened 2024-10-10
chore(codeowners): update code owners

martin308:

Having a valid code owner for the repository


beekhc (approved):
Nice catch! Would you mind doing the same change in the android sdk?

martin308:
@beekhc can do!

PR #19, opened 2024-10-18
feat(debug): Add configuration debug statements

arriIsHere:

setting `debug: true` in the config currently does nothing

Adds debug statements to the SDK, print out basic debugging information on SDK init.

1. set debug to true and run the smoke test/example
2. debug information should print out in the console.

### DIFF
--- a/Sources/Honeycomb/Honeycomb.swift
+++ b/Sources/Honeycomb/Honeycomb.swift
@@ -39,6 +39,11 @@ public class Honeycomb {
     static private let metricKitSubscriber = MetricKitSubscriber()
 
     static public func configure(options: HoneycombOptions) throws {
+
+        if options.debug {
+            configureDebug(options: options)
+        }
+
         guard let tracesEndpoint = URL(string: options.tracesEndpoint) else {
             throw HoneycombOptionsError.malformedURL(options.tracesEndpoint)
         }
@@ -167,10 +172,6 @@ public class Honeycomb {
 
         if #available(iOS 13.0, macOS 12.0, *) {
             MXMetricManager.shared.add(self.metricKitSubscriber)
-        }
-
-        if options.debug {
-            debugOptions(options: options)
         }
     }
 }
--- a/Sources/Honeycomb/Honeycomb.swift
+++ b/Sources/Honeycomb/Honeycomb.swift
@@ -168,5 +168,9 @@ public class Honeycomb {
         if #available(iOS 13.0, macOS 12.0, *) {
             MXMetricManager.shared.add(self.metricKitSubscriber)
         }
+
+        if options.debug {
+            debugOptions(options: options)
+        }
     }
 }
--- a/Sources/Honeycomb/HoneycombDebug.swift
+++ b/Sources/Honeycomb/HoneycombDebug.swift
@@ -3,7 +3,7 @@ import Foundation
 internal func configureDebug(options: HoneycombOptions) {
 
     if options.debug {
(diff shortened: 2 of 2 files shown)


arriIsHere:
I thought about adding tests but TBH I don't really see much of a use in them for debug statements. Really all they would test was if the swift logging system worked or if we can do boolean conditionals if I compared log statements.

Sources/Honeycomb/HoneycombDebug.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+internal func debugOptions(options: HoneycombOptions) {
```
arriIsHere, line 3: I used the JS library as my base for debug messaging and I didn't see a lot of places where we printed debug statements aside from the config options.

I can think of a few other places such as the sampler and span processor that we could add statements if we think it's useful. Not sure what would be printed though.

beekhc: I think there's some value in making it consistent with the web SDK, although I am a little curious why the web SDK is so selective about what it prints. I would've expected it to basically just print the options map as-is. 🤔 But that's fine. We can start with this, and if we want to change it, we can change both.

I would suggest renaming the method to `configureDebug` to be even more consistent with the web SDK. I think the name `debugOptions` is kinda confusing.

Also, completely optional, but I think it would be okay to put the 🐝 emoji in the first string like web has, if you want to. Apple is pretty good about emoji support everywhere.

Sources/Honeycomb/Honeycomb.swift
```diff
@@ -168,5 +168,9 @@ public class Honeycomb {
         if #available(iOS 13.0, macOS 12.0, *) {
             MXMetricManager.shared.add(self.metricKitSubscriber)
         }
+
+        if options.debug {
```
beekhc, line 172: Would it make more sense to put this at the top of the function, in case the options cause some kind of error, or if we later add some other debug lines during configuration?

beekhc (changes requested):
Nice. I'm glad we didn't end up doing one of the hackier options we discussed. Just a couple of small suggestions.

arriIsHere:
> > I thought about adding tests but TBH I don't really see much of a use in them for debug statements. Really all they would test was if the swift logging system worked or if we can do boolean conditionals if I compared log statements.
> 
> I agree. We chose not to add tests for Android either, because the difficultly of doing the plumbing for the tests was more work than the risk of messing up debug output would justify. We can always add debug tests later if we decide we need them.
> 
> I would ask that you run smoke tests with debug set to both true and false before you submit. We didn't do that for Android, and we ended up breaking CI. The Swift CI runs the smoke tests now, so that shouldn't happen. But I'd rather err on the safe side this time.

I ran smoke tests locally and seem to work well. It appears we do have debug enabled for the CI ones as well.

Sources/Honeycomb/HoneycombDebug.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 internal func debugOptions(options: HoneycombOptions) {
 
     if options.debug {
-        print("Honeycomb SDK Debug Mode Enabled")
+        print("🐝 Honeycomb SDK Debug Mode Enabled🐝 ")
```
beekhc, line 6: tiny nit: the space after the bee should be before the bee :-)

beekhc (approved):
nice! 🚀

Sources/Honeycomb/HoneycombDebug.swift
beekhc: I mean, on the second bee

PR #20, opened 2024-10-30
feat: SwiftUI View Rendering Instrumentation

MustafaHaddara:

- Closes #<enter issue here>

SwiftUI rendering is very interesting! 

This PR currently has a whole bunch of instrumentation and logging which makes it easier to see what's going on

This PR intentionally does not ship a view modifier form of the instrumentation. This is because ViewModifiers receive a `Content` struct (vs. Views that receive an `@ViewBuilder () -> Content` callback!); that is: ViewModifiers receive a view _after it has been initialized_, and therefore have no way to retroactively start a span before the view begins initializing. 

Compare the spans/log output from using the wrapper view form:

image

with the view modifier form:
image

And, now that we've settled on the wrapper form, here's the kinds of traces we're producing:

image


beekhc (commented):
Some early feedback.

Examples/SmokeTest/SmokeTest/ContentView.swift
```diff
@@ -33,32 +33,46 @@ private func flush() {
...
+                    Text("Send fake MetricKit data")
+                }
+                .buttonStyle(.bordered)
+
+                Button(action: flush) {
+                    Text("Flush")
+                }
+                .buttonStyle(.bordered)
+
+                HoneycombInstrumentedView(name: "expensive text") {
```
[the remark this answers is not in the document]
MustafaHaddara: I'm thinking I'll put this on a separate tab, like you did for network instrumentation, and add some controls or sliders or something to control how painful it is

Examples/SmokeTest/SmokeTest/ContentView.swift
```diff
@@ -65,6 +65,10 @@ struct ContentView: View {
             NetworkView()
                 .padding()
                 .tabItem { Label("Network", systemImage: "network") }
+
+            ViewInstrumentationView()
+                .padding()
+                .tabItem { Label("View Instrumentation", systemImage: "ruler") }
```
beekhc, line 71: It's really hard to resist the urge to bikeshed what icon to use for each tab 😆

Examples/SmokeTest/SmokeTest/ViewInstrumentationView.swift
```diff
@@ -0,0 +1,98 @@
...
+                ) {
+                    Text("Delay")
+                } minimumValueLabel: {
+                    Text("0")
+                } maximumValueLabel: {
+                    Text("4")
+                } onEditingChanged: { editing in
+                    isEditing = editing
+                    if !editing {
+                        delay = sliderDelay
```
beekhc, line 43: I really thought that since the value of the slider was a `$` binding that you would just be able to have `Slider(value: $delay,` and not have any `onEditingChanged:` handler. If this is necessary, could you add a comment explaining why we need all three state variables?

Examples/SmokeTest/SmokeTest/ViewInstrumentationView.swift
```diff
@@ -0,0 +1,98 @@
...
+        }
+        .onDisappear {
+            isEnabled = false
+        }
+    }
+}
+
+private func timeConsumingCalculation(_ delay: Double) -> String {
+    print("starting time consuming calculation")
+    sleep(UInt32(delay))
```
beekhc, line 92: `sleep()` is the POSIX function that takes a whole number of seconds, but it would be better to use the Swift function that takes a `TimeInterval`: `Thread.sleep(forTimeInterval: delay)`

Examples/SmokeTest/SmokeTest/ViewInstrumentationView.swift
```diff
@@ -0,0 +1,98 @@
...
+                ExpensiveView()
+            }
+        }
+        .onDisappear {
+            isEnabled = false
+        }
+    }
+}
+
+private func timeConsumingCalculation(_ delay: Double) -> String {
```
beekhc, line 90: Rather than `Double`, it would be more idiomatic and self-documenting to use `TimeInterval`, which is an alias for `Double`. I would try to use it as much as possible. Like, if we still need two state variables above, probably at least the `delay` one should be `TimeInterval` too.

Sources/Honeycomb/HoneycombInstrumentedView.swift
```diff
@@ -0,0 +1,64 @@
...
+        self.span = getViewTracer().spanBuilder(spanName: "View Render")
+            .setStartTime(time: Date())
+            .setAttribute(key: "ViewName", value: name)
+            .startSpan()
+
+        print("\(name) init")
+    }
+
+    var body: some View {
+        print("\(name) body started rendering")
```
beekhc, line 26: Don't forget to remove `print` statements.

Sources/Honeycomb/HoneycombInstrumentedView.swift
```diff
@@ -0,0 +1,64 @@
...
+            key: "RenderDuration",
+            value: endTime.timeIntervalSince(start)
+        )
+
+        return c.onAppear {
+            // contents end render
+
+            print("\(name) content appeared")
+            span.setAttribute(key: "TotalDuration", value: Date().timeIntervalSince(initTime))
+            span.end(time: Date())
```
beekhc, line 53: Would it be better to capture the `Date()` once and use the same value on both these lines?

Sources/Honeycomb/HoneycombInstrumentedView.swift
```diff
@@ -0,0 +1,64 @@
...
+            bodySpan.end(time: endTime)
```
beekhc, line 54: I'm a little confused about this. Is there a reason we can't do this on line 40, right after `endTime` is set?

beekhc (commented):
Overall, this looks great, and I just have a bunch of nits. But also, are you going to add some SmokeTest UI tests and assertions?

Examples/SmokeTest/SmokeTest/ViewInstrumentationView.swift
MustafaHaddara: yeah the idea was that we'd be able to adjust the slider but the rest of the view wouldn't re-render until you stopped editing it. 

BUT clearly I wasn't paying enough attention and that wasn't actually working. I'll restructure it and add a comment.

Sources/Honeycomb/HoneycombInstrumentedView.swift
MustafaHaddara: I wanted it to remain active in context, I put some comments in the code that should it explain it better.

Examples/SmokeTest/SmokeTest/ViewInstrumentationView.swift
MustafaHaddara: works now:

https://github.com/user-attachments/assets/eb5e5d50-e45d-4698-a9d9-f158c90183ac

(yes i know the simulator has a screen recorder but i wanted to capture my mouse cursor lol)

Examples/SmokeTest/SmokeTest/ViewInstrumentationView.swift
MustafaHaddara: yeah the idea was that we'd be able to adjust the slider but the rest of the view wouldn't re-render until you stopped editing it. 

BUT clearly I wasn't paying enough attention and that wasn't actually working. I'll restructure it and add a comment.

MustafaHaddara: works now:

https://github.com/user-attachments/assets/eb5e5d50-e45d-4698-a9d9-f158c90183ac

(yes i know the simulator has a screen recorder but i wanted to capture my mouse cursor lol)

smoke-tests/test_helpers/utilities.bash
```diff
@@ -20,7 +20,7 @@ span_attributes_for() {
 #   $3 - attribute key
 #   $4 - attribute type
 attribute_for_span_key() {
-	attributes_from_span_named $1 $2 | \
+	attributes_from_span_named "$1" "$2" | \
```
MustafaHaddara, line 23: We're emitting spans named `View Render` so these args need to be quoted for our utils to work correctly.

beekhc (approved):
This is great. I made a few comments for future discussion, but they shouldn't prevent us from submitting this.

PR #21, opened 2024-11-05
feat: Implement URLSession auto-instrumentation.

beekhc:

Adds auto-instrumentation for URLSession-based networking in iOS.

* Implement URLSession auto-instrumentation using swizzling.
* Add a "Network Playground" to the SmokeTest app for exercising networking API.
* Add smoke tests to make sure all the network requests get instrumented.
* Add BaggageSpanProcessor to tracer setup like our other SDKs, since we need that for tests.
* Remove unused duplicate file `smoke-tests/docker-compose.yml`.
* Add a mock-server for testing http requests against, the same as our Android SDK.
* Add a config file for how the test scheme is run, so that it won't run in parallel.

`make smoke`

Or run the example app, click the buttons, and look at collector logs.


Sources/Honeycomb/Networking/URLSessionTask.swift
```diff
@@ -0,0 +1,51 @@
+import Foundation
+import OpenTelemetryApi
+
+/// Returns true if this HTTP request is from our SDK itself, so that we don't recursively capture
+/// our own requests.
+private func isOTLPRequest(_ request: URLRequest) -> Bool {
+    // Just check for the OTLP version header that's always set.
+    if let headers = request.allHTTPHeaderFields {
+        for (key, _) in headers {
+            if key == "x-otlp-version" {
```
martin308, line 10: should we check other headers as well? honeycomb api key header or something?

Sources/Honeycomb/Networking/NetworkInstrumentation.swift
```diff
@@ -0,0 +1,49 @@
...
+internal func createSpan(from request: URLRequest) -> any Span {
+    let tracer = OpenTelemetry.instance.tracerProvider.get(
+        instrumentationName: urlSessionInstrumentationName,
+        instrumentationVersion: honeycombLibraryVersion
+    )
+
+    var builder = tracer.spanBuilder(spanName: request.httpMethod ?? "UNKNOWN")
+    builder.setSpanKind(spanKind: SpanKind.client)
+    if let method = request.httpMethod {
+        builder = builder.setAttribute(key: "http.request.method", value: method)
```
martin308, line 18: can/should we use the semantic attributes from the `OpenTelemetryApi` module?

Sources/Honeycomb/Honeycomb.swift
```diff
@@ -96,8 +96,11 @@ public class Honeycomb {
             }
         let spanProcessor = BatchSpanProcessor(spanExporter: spanExporter)
 
+        let baggageSpanProcessor = HoneycombBaggageSpanProcessor(filter: { _ in true })
+
         let tracerProvider = TracerProviderBuilder()
             .add(spanProcessor: spanProcessor)
+            .add(spanProcessor: baggageSpanProcessor)
```
martin308, line 103: is this part of the URLSession instrumentation?

martin308 (approved):

Sources/Honeycomb/Networking/URLSessionTask.swift
beekhc: This was pretty ad-hoc, and I'm open to other suggestions, but I think it's probably best to keep it simple. I don't think we want to require the honeycomb api key, because it's not strictly necessary if the app is talking to a collector, right?

This is the only required header listed on our SDK distro spec
https://github.com/honeycombio/specs/blob/main/specs/otel-sdk-distro.md#headers

Sources/Honeycomb/Networking/NetworkInstrumentation.swift
beekhc: Done.

Thanks, this is much nicer. Interestingly, I had to switch to setting them on the `Span`, because that setter wasn't present on the `SpanBuilder`. I think maybe that's an oversight we could contribute upstream? I'll file a ticket.

Sources/Honeycomb/Honeycomb.swift
beekhc: Sort of. It's something we do in our other SDKs, and by adding it here, I can more easily to implement the smoke tests, by setting baggage before I make the request, and then checking that those ended up on the spans.

I could pull it out into its own PR, but since it's so small and should be uncontroversial, I was just gonna do it here. But if you feel strongly, I can pull it out.

Sources/Honeycomb/Networking/URLSessionTask.swift
martin308: true, I forgot about the collector scenario. Makes sense 👍

beekhc:
I didn't realize that running the mock server on circleci would be an issue. I'm gonna need to look into that.

.circleci/config.yml
```diff
@@ -64,6 +64,10 @@ jobs:
           name: Start OTel Collector
           command: touch ./smoke-tests/collector/data.json && ./otelcol --config=./smoke-tests/collector/otel-collector-config.yaml --set="exporters::file::path=./smoke-tests/collector/data.json"
           background: true
+      - run:
+          name: Start Mock Server
+          command: cd smoke-tests/mock-server && npm install && npm start
+          background: true
```
martin308, line 70: you could also set working directory for this I think if you wanted

```suggestion
      - run:
          name: Start Mock Server
          working_directory: smoke-tests/mock-server
          command: npm install && npm start
          background: true
```

martin308 (approved):

MustafaHaddara (approved):

PR #22, opened 2024-11-15
feat: UIKit Navigation instrimentation

arriIsHere:

Adds auto-instrimentation for UIKit navigation.

- Closes #<enter issue here>

Swizzles the UIViewController `viewDidAppear` and `viewDidDisappear` to observe 

Some smoke tests were added that verify the collection of data (WIP needs to be more granular)

### DIFF
--- a/Examples/SmokeTest/SmokeTest/ContentView.swift
+++ b/Examples/SmokeTest/SmokeTest/ContentView.swift
@@ -69,8 +69,8 @@ struct ContentView: View {
             ViewInstrumentationView()
                 .padding()
                 .tabItem { Label("View Instrumentation", systemImage: "ruler") }
-            
-            UIKView()
+
+            UIKitView()
                 .padding()
                 .tabItem {
                     Label(
--- a/Examples/SmokeTest/SmokeTest/ContentView.swift
+++ b/Examples/SmokeTest/SmokeTest/ContentView.swift
@@ -69,6 +69,11 @@ struct ContentView: View {
             ViewInstrumentationView()
                 .padding()
                 .tabItem { Label("View Instrumentation", systemImage: "ruler") }
+            
+            UIKView()
+                .padding()
+                .tabItem { Label("UIKit",
+                    systemImage: "paintpalette" )}
         }
     }
 }
--- a/Examples/SmokeTest/SmokeTest/ContentView.swift
+++ b/Examples/SmokeTest/SmokeTest/ContentView.swift
@@ -72,8 +72,12 @@ struct ContentView: View {
             
             UIKView()
                 .padding()
-                .tabItem { Label("UIKit",
-                    systemImage: "paintpalette" )}
+                .tabItem {
+                    Label(
+                        "UIKit",
+                        systemImage: "paintpalette"
+                    )
+                }
         }
     }
 }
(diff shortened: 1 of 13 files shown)


Examples/SmokeTest/SmokeTest/ContentView.swift
```diff
@@ -65,6 +65,15 @@ struct ContentView: View {
             NetworkView()
                 .padding()
                 .tabItem { Label("Network", systemImage: "network") }
+
+            UIKView()
```
beekhc, line 69: The name `UIKView` feels very wrong to me. Objective C didn't have namespaces, so the convention was to use two letter (and later, three letter) prefixes to avoid naming collisions. "UI" was the prefix for UIKit. My brain can't help but read `UIKView` as a `View` in the `UIK` namespace. I'd recommend just making it `UIKitView`.

Examples/SmokeTest/SmokeTestUITests/SmokeTestUITests.swift
```diff
@@ -153,6 +153,17 @@ final class SmokeTestUITests: XCTestCase {
         app.buttons["Flush"].tap()
     }
 
+    func testUIKitInstrimentation() throws {
+        let app = XCUIApplication()
+        app.launch()
+        app.buttons["UIKit"].tap()
+        XCTAssert(app.staticTexts["Sample UIKit App"].waitForExistence(timeout: uiUpdateTimeout))
```
beekhc, line 160: Since we're swizzling `UIViewController`, this instrumentation should work for both `UINavigationContoller` and `UITabBarController`, among others. Should we add tests for both a navigation controller and a tab view, to make sure they're both reasonable? I'm not sure how annoying that would be to build in Storyboard.

Sources/Honeycomb/UIKit/NavigationInstrumentation.swift
```diff
@@ -0,0 +1,18 @@
+#if canImport(UIKit)
+    import Foundation
+    import OpenTelemetryApi
+    import UIKit
+
+    private let honeycombInstrumentationView = "@honeycombio/instrumented-view"
```
beekhc, line 6: PR #20 is probably gonna go in before this one, and has `"@honeycombio/instrumentation-view"`. Let's find a place where we can put it as a constant and reuse it. Maybe all of our tracer names should be in one file? 🤔

beekhc (changes requested):
Just wanted to give some early feedback, even though I know we're discussing some of the details later.

I think we should also document the new spans and attributes in the Readme.

Examples/SmokeTest/SmokeTest/UIKitView.swift
```diff
@@ -0,0 +1,42 @@
+//
+//  UIKView.swift
+//  SmokeTest
+//
+//  Created by Arri Blais on 11/13/24.
+//
```
beekhc, line 6: You should delete these headers. You can change the template in Xcode settings somewhere.

README.md
```diff
@@ -76,6 +76,16 @@ To manually send a span:
 The following auto-instrumentation libraries are automatically included:
 * [MetricKit](https://developer.apple.com/documentation/metrickit) data is automatically collected.
 
+### UIKit instrumentation
```
beekhc, line 79: nit: It looks like we capitalized "Instrumentation" everywhere else, so let's be consistent.

beekhc (approved):
This is great! 🚀

PR #23, opened 2024-11-15
feat: instrumented NavigationStack in SwiftUI

MustafaHaddara:

Provides helpers for instrumentating navigations

Exposes 2 APIs:

- A ViewModifier: `.instrumentNavigations(path: [any])` or `.instrumentNavigations(path: NavigationPath)`
- A helper function: `reportNavigation(path: any)`

(I'm not overly attached to this interface so definitely let me know if we think we should change it. ex. I could see us shipping a `reportNavigationOnAppear()` view modifier to encapsulate the `onAppear { reportNavigation() }` pattern I have going, or we could change the names of some of these, or whatever...)

Both emit the same types of the spans and also set the current path in global state. This PR also sets up a SpanProcessor that reads that global state and includes the current path as an attribute on the span. 

(More info and full documentation included in the changes to `README.md`)

- [x] smoke tests
- [x] run the sample app, click around, watch the navigation spans come through in the collector/stdout/your honeycomb account


Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
```diff
@@ -0,0 +1,90 @@
...
+var currentNavigationPath: String? = nil
+
+func reportNavigation(path: Encodable) {
+    do {
+        let encoder = JSONEncoder()
+        let data = try encoder.encode(path)
+        let pathStr = String(decoding: data, as: UTF8.self)
+        print("current path: \(pathStr)")
+
+        currentNavigationPath = pathStr
```
MustafaHaddara, line 40: Ok this looks super dumb.

BUT! The "correct" way to do this (with baggage or context) seems busted. 
- otel swift doesn't seem to expose an open "write whatever you want into context" method (and if it does, i couldn't find it)
- otel swift DOES offer a way to dumb arbitrary data into baggage, but baggage is immutable, meaning every new value in baggage needs to also call setActiveBaggage(), meaning that some other code could come along and randomly decide to nuke it without us being able to do anything about it.

ie. instead of just assigning to this global variable, I was doing this:

```swift
let baggage = OpenTelemetry.instance.baggageManager.baggageBuilder()
    .put(key: EntryKey("CurrentNavigationPath")!, value: EntryValue(string: pathStr)!, metadata: nil)
    .build()

OpenTelemetry.instance.contextProvider.setActiveBaggage(baggage)
```

Then I was relying on our BaggageSpanProcessor to pull the value out of baggage and stick it on each span's attributes. However, this worked for some spans and then for others, they just wouldn't get the right attribute. As far as I could tell, something in the process was clearing out the active Baggage, but I couldn't figure out what was doing it.

smoke-tests/smoke-e2e.bats
```diff
@@ -200,5 +201,14 @@ mk_diag_attr() {
...
+    assert_equal "$result" '   6 "SwiftUI.UIKitTabBarController"'
+}
+
+@test "Navigation attributes are correct" {
+    result=$(attribute_for_span_key "@honeycombio/instrumentation-navigation" Navigation CurrentNavigationPath string | sort | uniq -c)
+    assert_equal "$result" '  10 "Core"
+   1 "Network"
+   1 "View Instrumentation"
+   1 "[]"
+   1 "[{\"name\":\"Yosemite\"}]"'
```
MustafaHaddara, line 213: this feels fragile to me because the more we navigate around in our tests, the more spans we'll get here. Perhaps I should be filtering it down and checking that `"[]"` and `"[{\"name\":\"Yosemite\"}]"'` exist at least once?

Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
beekhc: Is this something we should fix in upstream Otel? I agree it's annoying that baggage isn't mutable, and even if it were immutable, it seems like there should be an API to extend it with a new entry.

We could write a Baggage extension that copies the existing keys and adds new ones, and then work on upstreaming it later.

README.md
```diff
@@ -114,3 +114,92 @@ Specifically, it will emit 2 kinds of span for each view that is wrapped:
...
+            .navigationDestination(for: Park.self) { park in
+                ParkDetails(park: park)
+            }
+        }
+        .instrumentNavigations(path: presentedParks) // convenience View Modifier
+    }
+}
+```  
+
+Whenever the `path` variable changes, this View Modifier will emit a span with the name `Navigation`. This span will contain the following attributes:
```
beekhc, line 144: We should probably make these names consistent with the UIKit navigation instrumentation, right? I'm worried that all our different trace names look different.

Examples/SmokeTest/SmokeTest/NavigationExamplesView.swift
```diff
@@ -0,0 +1,137 @@
...
+    var body: some View {
+        NavigationStack(path: $presentedParks) {
+            List(parks) { park in
+                NavigationLink(park.name, value: park)
+            }
+            .navigationDestination(for: Park.self) { park in
+                ParkDetails(park: park)
+            }
+        }
+        .instrumentNavigations(path: presentedParks)
```
beekhc, line 77: nit: I think `instrumentNavigations` sounds awkward. Maybe just drop the "s" on the end?

Examples/SmokeTest/SmokeTest/NavigationExamplesView.swift
```diff
@@ -0,0 +1,137 @@
...
+    @State private var selectedPark: Park.ID? = nil
+    @State private var selectedTree: Park.ID? = nil
+
+    var body: some View {
+        NavigationSplitView {
+            List(parks, selection: $selectedPark) { park in
+                Text(park.name)
+            }
+            .onAppear {
+                reportNavigation(path: "Split View Parks Root")
```
beekhc, line 91: As a shape, this seems fine to me? I hear what you're saying about adding a helper that does both the `onAppear` and the `reportNavigation`, but would that cause problems if they add their own `onAppear` before or after our modifier?

I'm also not sure about `reportNavigation` as a top-level name, because I'm worried it's too generic. I mean, like, Sentry has `.sentryTrace()` and our instrumented view name starts with `Honeycomb`. But `reportNavigationToHoneycomb` is pretty ridiculous... I dunno.

beekhc: We could put it on the `Honeycomb` object, so it would be `Honeycomb.reportNavigation`, but that feels wrong to me -- using a class as a namespace.

Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
MustafaHaddara: The problem isn't setting Baggage, it's convincing everything to use the same extension/methods. 

I was able to create new baggage with what I needed, but some other code somewhere else seemed to overwriting it :sob:

README.md
MustafaHaddara: ie. emitting `viewDidAppear` as the span name here?

Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
MustafaHaddara: circling back: based on internal discussion on slack, it sounds like baggage isn't even the best choice for this pattern. 

Right now I'm just using a global mutable variable, which feels silly and unconventional. Is there a better Swift-y conventional API that I should be using instead?

README.md
beekhc: I guess? Would they stomp on each other? I guess we'll find out from the smoke tests.

Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
beekhc: I think having a mutable global singleton is probably the best option, since that's how the OpenTelemetry SDK already exposes its APIs, and this is truly global state.

I wonder if we can solve both this and the naming problem with the same solution. Maybe if we have a singleton like `HoneycombNavigation` and call `HoneycombNavigation.setCurrentScreen()`, then we could record those as navigation events as well as keeping track of the state for a custom span processor to add the attributes?

I guess that's not any better than `Honeycomb.setCurrentScreen()` though. Well, I'm open to suggestions. But I kinda like the idea of having a straightforward "This is the screen you're on" API that logs the event (for Timeline Analysis) and also sets the global attribute. Then that API could work with _any_ UI framework, and we should even call it for the UIKit instrumentation, assuming they aren't constantly clobbering each other.

MustafaHaddara: I think I misread your previous suggestion about putting it on the global honeycomb object. I think that works best. 

I'll rewrite this into a shared singleton and then expose that from the main honeycomb object.

Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
```diff
@@ -6,84 +6,85 @@ private let navigationInstrumentationName = "@honeycombio/instrumentation-naviga
...
 }
 
-@available(iOS 16.0, macOS 12.0, *)
-func reportNavigation(path: NavigationPath) {
-    if let codablePath = path.codable {
-        reportNavigation(path: codablePath)
-    } else {
-        reportNavigation(path: unencodablePath)
-    }
+internal class HoneycombNavigationProcessor {
```
MustafaHaddara, line 16: I was expecting the `internal` keyword here to limit access to this class, but it looks like I could still call it from our smoke test app. 

ie. the following, in a view in the smoke test app, was **not** an error and compiled successfully:

```swift
HoneycombNavigationProcessor.singleton.reportNavigation(path: "asdf")
```

What am I missing in order to limit access to this singleton and force clients to go through the entrypoints in the main Honeycomb class?

Examples/SmokeTest/SmokeTest/ContentView.swift
```diff
@@ -78,6 +86,10 @@ struct ContentView: View {
                         systemImage: "paintpalette"
                     )
                 }
+
+            NavigationExamplesView()
+                .padding()
+                .tabItem { Label("Navigation", systemImage: "globe") }
```
beekhc, line 92: Do we want to add an `onAppear` to this one like the others?

Examples/SmokeTest/SmokeTest/NavigationExamplesView.swift
```diff
@@ -0,0 +1,136 @@
...
+    let park: Park
+    let tree: Tree
+    var body: some View {
+        Text("\(park) has many \(tree)")
+    }
+}
+
+struct NavigationStackExample: View {
+    @State private var presentedParks: [Park] = []
+    //  @State private var presentedParks = NavigationPath()
```
beekhc, line 65: Can we remove this commented out code?

Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
```diff
@@ -0,0 +1,115 @@
...
+
+func getTracer() -> Tracer {
+    return OpenTelemetry.instance.tracerProvider.get(
+        instrumentationName: navigationInstrumentationName,
+        instrumentationVersion: honeycombLibraryVersion
+    )
+}
+
+internal class HoneycombNavigationProcessor {
+    static let singleton = HoneycombNavigationProcessor()
```
beekhc, line 17: fyi, if this were a public API, I would recommend naming it either `instance` or `shared`, as those are more idiomatic, but for an internal instance, I don't think it matters.

Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
```diff
@@ -0,0 +1,115 @@
...
+    func reportNavigation(path: NavigationPath) {
+        if let codablePath = path.codable {
+            reportNavigation(path: codablePath)
+        } else {
+            reportNavigation(path: unencodablePath)
+        }
+    }
+
+    func reportNavigation(path: String) {
+        print("current path: \(path)")
```
beekhc, line 32: I think this was just for debug and can be removed?

Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
```diff
@@ -0,0 +1,115 @@
...
+    public let isStartRequired = true
+    public let isEndRequired = false
+
+    public func onStart(
+        parentContext: SpanContext?,
+        span: any ReadableSpan
+    ) {
+        if HoneycombNavigationProcessor.singleton.currentNavigationPath != nil {
+            span.setAttribute(
+                key: "CurrentNavigationPath",
```
beekhc, line 103: In Android, they call this attribute `screen.name`. I think it would be good to be consistent with that, especially for use in Timeline Analysis.

https://github.com/open-telemetry/opentelemetry-android/blob/c10adbaf11dbc415190d5fdf8dcf5183485d3ef6/core/src/main/java/io/opentelemetry/android/common/RumConstants.java#L20

smoke-tests/smoke-e2e.bats
beekhc: What I ended up with for the other PR was extracting a whole span, like this:

```
# Spans on a particular view.
 # Arguments:
 #   $1 - scope
 #   $2 - span name
 #   $3 - view.name
 spans_on_view_named() {
     spans_received | jq ".scopeSpans[] \
         | select(.scope.name == \"$1\").spans[] \
         | select (.name == \"$2\") as \$span \
         | .attributes?[]? \
         | select (.key? == \"view.name\" and .value.stringValue == \"$3\") \
         | \$span"
 }
```

And then individually picking out the attributes, like this:

```
    class=$(echo "$span" | jq '.attributes[] | select(.key == "view.class").value.stringValue')
    assert_equal "$class" '"UIButton"'
```

Maybe something similar would work here?

Examples/SmokeTest/SmokeTest/ContentView.swift
MustafaHaddara: whoops yeah. Love those logical merge conflicts.

MustafaHaddara: Aha! This is kinda subtle. The UIKit view did get added after the fact and needs an onAppear 

This one, though, doesn't need it, because, the nested NavigationStack / NavigationSplitView will do it for us.

Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
MustafaHaddara: I'll switch to `shared`

Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
MustafaHaddara: yeah

Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
MustafaHaddara: sounds good. Do you want me to change it on the Navigation spans as well?

beekhc: yeah, I think so

smoke-tests/smoke-e2e.bats
MustafaHaddara: I think for this particular test, all I want is "did the navigation spans get triggered when we navigate to the screen" and I can do that in a less fragile way by grepping for the navigation lines I want. 

But I can use this method on a different set of spans to ensure, ex. the render spans include the right attribute

beekhc (approved):

PR #24, opened 2024-11-19
fix: Auto-select the SDK when running smoke tests.

beekhc:

Currently, our smoke tests are hard-coded to a particular SDK we know is available in circleci. But individual developers have updated Xcode and no longer have that version. So, everyone has to edit this file to make the smoke tests run. This PR changes the script to autodetect the SDK version to use.

`xcodebuild -showsdks` lists the SDKs. We filter for `iphonesimulator`, and then use an ad-hoc regex to get just the name of the SDK.

I tested it locally. If it still passes in circleci, it should be good to go.


run-ios-tests.sh
```diff
@@ -1,6 +1,8 @@
 set -e
 
-SDK="iphonesimulator17.5"
+SDK=$(xcodebuild -showsdks | grep iphonesimulator | sed -e 's/^.*-sdk //')
+echo "SDK: $SDK"
```
MustafaHaddara, line 4: 😍

MustafaHaddara (approved):
ship it!

PR #27, opened 2025-01-13
fix: fix resource construction to include defaults

beekhc:

Our SDK was not setting `telemetry.sdk.language` and other default resource fields, even though the upstream SDK was.

We need to call the _default_ constructor for the `Resource` object and then merge in our own fields, rather than just passing in our own fields.

There is a new smoke test.


arriIsHere (approved):
🚀

PR #28, opened 2025-01-14
maint: various small fixes to core swift SDK

beekhc:

These are various small cleanups as we prepare for an eventual beta release.

* Updated CHANGELOG.md with all changes since last release.
* Updated OTel to 1.12.1.
* Updated `RELEASING.md` to better match this repo's process.
* Removed old TODOs from `Honeycomb.swift`. Some were fixed already, and others were deprioritized.
* Changed debug options to print _all_ options, rather than just a handful.
* Stopped hard-coding `otlpVersion` since that's fixed upstream.
* Removed a constructor that is now redundant for `HoneycombOptions`.

Smoke tests and unit tests still pass. Manually verified debug output is reasonable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 Honeycomb OpenTelemetry SDK Changelog
 
+## v.Next
+
+### New Features
+
+* Update to OpenTelemetry Swift 1.12.1.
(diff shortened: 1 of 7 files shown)


wolfgangcodes (approved):
🚢

PR #38, opened 2025-02-21
feat: expose `addSpanProcessor()` api for custom span processors

MustafaHaddara:
We need some API for clients to provide a custom span processor!

- Adds a `.setSpanProcessor()` method to `HoneycombOptions`
    - `HoneycombOptions` has a `configureFromSource` method that reads values out of a plist file, but I didn't update that. I wasn't sure how we'd pass in a full object with methods and such from a plist file. @beekhc let me know if you have ideas or if you think this is ok.
- Use that when setting up the SDK
- Adds an internal-only CompositeSpanProcessor that allows for multiple span processors
    - should we expose this as well? it might be useful

- [x] smoke tests pass

---

- [x] Changelog is updated

### DIFF
--- a/Sources/Honeycomb/CompositeSpanProcessor.swift
+++ b/Sources/Honeycomb/CompositeSpanProcessor.swift
@@ -6,29 +6,32 @@ internal class CompositeSpanProcessor: SpanProcessor {
     private var spanProcessors: [SpanProcessor] = []
     var isStartRequired: Bool = false
     var isEndRequired: Bool = false
-    
+
     func addSpanProcessor(_ spanProcessor: SpanProcessor) {
         spanProcessors.append(spanProcessor)
         isStartRequired = spanProcessors.contains(where: { $0.isStartRequired })
         isEndRequired = spanProcessors.contains(where: { $0.isEndRequired })
     }
 
-    func onStart(parentContext: OpenTelemetryApi.SpanContext?, span: any OpenTelemetrySdk.ReadableSpan) {
+    func onStart(
+        parentContext: OpenTelemetryApi.SpanContext?,
+        span: any OpenTelemetrySdk.ReadableSpan
+    ) {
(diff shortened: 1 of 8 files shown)


Sources/Honeycomb/CompositeSpanProcessor.swift
```diff
@@ -0,0 +1,40 @@
...
+import OpenTelemetrySdk
+
+internal class CompositeSpanProcessor: SpanProcessor {
+    private var spanProcessors: [SpanProcessor] = []
+    var isStartRequired: Bool = false
+    var isEndRequired: Bool = false
+
+    func addSpanProcessor(_ spanProcessor: SpanProcessor) {
+        spanProcessors.append(spanProcessor)
+        isStartRequired = spanProcessors.contains(where: { $0.isStartRequired })
```
arriIsHere, line 12: Would we need to do this over all the span processors everytime? does the "start/end required" change for already added ones?

arriIsHere (approved):
This looks good!

I just had one question about the `start/stop` required.

I wonder if we should maybe submit the `CompositeSpanProcessor` upstream?

Sources/Honeycomb/CompositeSpanProcessor.swift
MustafaHaddara: Theoretically, I guess it's possible for `isStartRequired`/`isEndRequired` to change at runtime, since we don't know what SpanProcessors we'll be receiving...

so maybe it makes more sense for these to become computed properties? Like so:

```swift
var isStartRequired: Bool {
    spanProcessors.contains(where: { $0.isStartRequired })
}
var isEndRequired: Bool {
    spanProcessors.contains(where: { $0.isEndRequired })
}
```

MustafaHaddara: ok I thought about it a bit more and I'm going to take the computed properties approach!

PR #35, opened 2025-02-18
feat: Add cocopods frameworks for publishing

arriIsHere:
## Which problem is this PR solving?

- Closes #<enter issue here>

## Short description of the changes

Adds coccoapods `podspec` file to allow us to publish this on the cocoapods repo. 

This also conditionally disables gRPC protocol if the grpc libraries are unavailable. This was necessary as there is no official `grpc-swift` package available on cocoapods upon which upstream otel-swift depends.

## TODO:

- Publishing this to cocoapods (preferably as part of release)
- ~~Upstream changes merged (so that people can download the otel dependencies).~~

---

- [X] Changelog is updated

PR #40, opened 2025-02-26
chore: use reverse url notation for instrumentations

MustafaHaddara:
`@honeycombio/` naming is a web convention. This PR switches it up to reverse url notation.

- Changed instrumentation names. I went with `io.honeycomb.instrumentation.*` notation, where the last chunk is a `kebab-case` name for the instrumentation.
- Changed smoke tests to match

Smoke tests still pass.

---

- [x] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ Honeycomb OpenTelemetry SDK Changelog
 
 ## v.Next
 
+* Update instrumentation names to use reverse url notation (`io.honeycomb.instrumentation.*`) instead of `@honeycombio/instrumentation-*` notation. 
+
 ## 0.0.5-alpha
 
 * Add a `setSpanProcessor()` function to `HoneycombOptions` builder to allow clients to supply custom span processors.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ Honeycomb OpenTelemetry SDK Changelog
 
 ## v.Next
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -2,9 +2,14 @@ Honeycomb OpenTelemetry SDK Changelog
...
 
+* Update instrumentation names to use reverse url notation (`io.honeycomb.instrumentation.*`) instead of `@honeycombio/instrumentation-*` notation. 
+* Package is now available on Cocoapods.
+
+## 0.0.5-alpha
+
 ### New Features
 
-* Package is now available on Cocoapods.
+* Add a `setSpanProcessor()` function to `HoneycombOptions` builder to allow clients to supply custom span processors.
```
MustafaHaddara, line 12: this restores the changelog that got wiped out in #35 😳

Sources/Honeycomb/MetricKitSubscriber.swift
```diff
@@ -2,7 +2,7 @@ import Foundation
 import MetricKit
 import OpenTelemetryApi
 
-private let metricKitInstrumentationName = "@honeycombio/instrumentation-metric-kit"
+private let metricKitInstrumentationName = "io.honeycomb.instrumentation.metric-kit"
```
beekhc, line 5: nit: if we're using `urlsession` and `uikit`, we should use `metrickit` here

beekhc (approved):

MustafaHaddara:
as we discussed on slack, i dropped the `.instrumentation` chunk from the lib names. We'll now use `io.honeycomb.*` instead of `io.honeycomb.instrumentation.*`

beekhc (approved):

PR #42, opened 2025-02-27
feat: add api to enable or disable instrumentation

beekhc:

Currently there is no way to choose which auto-instrumentation is enabled or disabled, but some consumers may not want all of them, especially the touch instrumentation, which is pretty noisy.

* Adds options for each of the current auto-instrumentations included in our SDK.
* Changes the touch instrumentation to default to off, since it is so noisy.
* Preëmptively adds an option for the upcoming unhandled exception auto-instrumentation.

* Adds new unit tests to verify the options get set.
* Updates the smoke tests to enable the touch instrumentation.

---

- [✅] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,13 @@ Honeycomb OpenTelemetry SDK Changelog
 
 ### New Features
 
+* Package is now available on Cocoapods.
 * Add new options to enable/disable built-in auto-instrumentation.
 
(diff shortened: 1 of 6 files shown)


MustafaHaddara (approved):
LGTM 👍 

We should probably update the readme
1. to call out that touch instrumentation defaults to False
2. to include the new options in the big table of options

PR #43, opened 2025-02-28
feat: Manual Error logging API

arriIsHere:

- Closes #<enter issue here>

Adds several new methods that will log errors of the types NSError, NSException, and Error (swift type). Each of the methods optionally takes attribute values, a thread (for calling thread) and a logger which will default to the `io.honeycomb.error` logger.

- Smoke tests have been added that verify the correctness and existence of log events.

---

- [X] Changelog is updated


Examples/SmokeTest/SmokeTest/ContentView.swift
```diff
@@ -3,6 +3,10 @@ import OpenTelemetryApi
 import OpenTelemetrySdk
 import SwiftUI
 
+enum TestErrors: Error {
```
beekhc, line 6: nit: `TestError`, not plural

Examples/SmokeTest/SmokeTest/ErrorHelpers/CatchNSException.m
```diff
@@ -0,0 +1,25 @@
+//
```
beekhc, line 1: i think we usually delete these headers?

Examples/SmokeTest/SmokeTest/ErrorHelpers/CatchNSException.m
```diff
@@ -0,0 +1,25 @@
...
+//  Created by Arri Blais on 2/26/25.
+//
+
+#import "CatchNSException.h"
+
+@implementation CatchNSException
+
++ (NSException *)throwAndCatchNSException {
+    NSException * exception = nil;
```
beekhc, line 13: FYI: rather than nitpicking the formatting here, I created a task for use to add an objc auto-formatter like we use for swift. I don't think this is super urgent, given how little objc code we have.

https://app.asana.com/0/1207614375503485/1209539503376007

Sources/Honeycomb/Honeycomb.swift
```diff
@@ -206,6 +206,101 @@ public class Honeycomb {
         }
     }
 
+    private static let errorLoggerInstrumentationName = "io.honeycomb.error"
+
+    public static let defaultErrorLogger = OpenTelemetry.instance.loggerProvider.get(
```
beekhc, line 211: I'm a little shocked that this works, since in most languages, static initialization of the `Honeycomb` class would have to happen before `Honeycomb.configure` is called. But I just looked it up and learned that Swift doesn't initialize static members until they are actually used 🤯 .

Still, it makes me nervous. But it's used in multiple places, so... I dunno. I guess it's fine.

Sources/Honeycomb/Honeycomb.swift
```diff
@@ -206,6 +206,101 @@ public class Honeycomb {
...
+    ) {
+        let timestamp = Date()
+        let type = String(describing: Mirror(reflecting: exception).subjectType)
+        let message = exception.reason ?? exception.name.rawValue
+
+        var errorAttributes = [
+            "exception.type": type.attributeValue(),
+            "exception.message": message.attributeValue(),
+            "exception.name": exception.name.rawValue.attributeValue(),
+            "exception.stacktrace": exception.callStackSymbols.attributeValue(),
```
beekhc, line 254: I think with all of our other error APIs, we store the callstack as one string, not an array of strings. We could just join the array on "\n" here... For MetricKit, I cowardly punted instead of figuring it out.

We could check with error folks and see if they have a recommendation, but I would err on the side of being consistent with the other SDKs.

smoke-tests/smoke-e2e.bats
```diff
@@ -318,3 +318,39 @@ mk_diag_attr() {
 @test "Span Processor gets added correctly" {
     result=$(spans_received | jq ".attributes[] | select (.key == \"app.metadata\").value.stringValue" "app.metadata" string | uniq)
 }
+
+@test "NSException attributes are correct" {
+    attrs=$(attributes_for_log_with_value "io.honeycomb.error" "NSException" string)
+
+    stacktrace=$(echo "$attrs" | jq "select (.key == \"exception.stacktrace\").value | .arrayValue.values[]")
```
beekhc, line 325: This is fine, but there are enough of these, I would consider adding a helper like `spans_on_view_named` that will capture the attrs using `jq`'s `as` feature to capture the attributes. I would probably just spend like 15 or 20 minutes seeing if there's a way to simplify the assertions, but if not, then leave it.

https://github.com/honeycombio/honeycomb-opentelemetry-swift/blob/81c1fe9cbd5ac1fd48e73fd76525459d9d4fba66/smoke-tests/test_helpers/utilities.bash#L8

beekhc (commented):
This is great. My only significant question is whether we really want the stack trace to be an array, vs just joining the string with "\n".

Sources/Honeycomb/Honeycomb.swift
arriIsHere: I can pull it out to a method so that we can ensure it's not sketchy.
Actually, yah I'll do that while I'm working on the PR comments this morning.

Sources/Honeycomb/Honeycomb.swift
beekhc: Makes sense.

beekhc (approved):
This is awesome! 🚀

PR #44, opened 2025-02-28
fix: make sessions threadsafe

beekhc:

We've noticed some strange issues with session IDs changing more frequently than they should. One theory is that there is a race condition, since session ID generation wasn't threadsafe. This PR makes it threadsafe.

Smoke tests sill pass. It's difficult to test that any potential race condition is fixed.

---

- [✅] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,18 +5,12 @@ Honeycomb OpenTelemetry SDK Changelog
 ### New Features
 
 * Error logging API for manually logging exceptions.
-
-### Fixes
-
-* Update instrumentation names to use reverse url notation (`io.honeycomb.*`) instead of `@honeycombio/instrumentation-*` notation. 
-
-### New Features
-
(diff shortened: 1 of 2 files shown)


martin308 (approved):
can we make the lock any more granular? 

Also NSLock in swift? 😂

beekhc:
> can we make the lock any more granular?

Not really. All the logic for checking expiration and making sure the ID is not null needs to be in the critical section. And at least we're not doing anything expensive while holding the lock.

> Also NSLock in swift? 😂

Yeah... I think Swift would prefer if we did all span processing on the same thread, but I don't _think_ OTel works that way. Swift did add a `Mutex` type eventually, but it's not supported on the oldest iOS versions we support.

Anyway, I'm just proposing this because it's simple and easy. If this doesn't solve the problem with changing session IDs too soon, we'll want to dig in and understand it more deeply.

PR #45, opened 2025-03-04
feat: Catch unhandled exceptions

arriIsHere:

Adds unhandled exception reporting and crash reporting to auto instrumentation. 

- Closes #<enter issue here>

UncaughtExceptionHandler and signal handers have been added to allow crash-litics

- Open the Smoke test application in xcode
- remove the catch statement from the NSException Objc code.
- Run the app and press the "Send fake error data" button.
- A crash should be logged

---

- [x] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Honeycomb OpenTelemetry SDK Changelog
 * Error logging API for manually logging exceptions.
 * Package is now available on Cocoapods.
 * Add new options to enable/disable built-in auto-instrumentation.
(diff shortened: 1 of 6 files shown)


arriIsHere:
I was having some trouble figuring out how I could test this.

beekhc (changes requested):
i don't think we need automated tests, but I do think it's worth adding a button to the smoke test app that causes one of these crashes and manually verifying it. that's what i did for android.

beekhc (approved):
Looks good! 🚀

PR #46, opened 2025-03-05
feat: offline support for telemetry

MustafaHaddara:
See also: https://github.com/honeycombio/honeycomb-opentelemetry-android/pull/50

Mobile apps may have unreliable network connections. Telemetry emitted while a mobile device is offline is important, and should not be lost.

Uses OpenTelemetry's (extremely alpha) `PersistenceExporter` library to add a local caching layer to our exporters. 

- Run the smoke test app, open the collector logs side by side
- Kill the collector
- Interact with the app
- Wait, Observe no telemetry gets sent.
    - I saw a `Error Domain=NSURLErrorDomain Code=-1004 "Could not connect to the server."` in the Xcode console.
- Bring up the collector again
- Observe the cached telemetry gets sent to the collector.

---

- [x] Changelog is updated


Sources/Honeycomb/PersistenceExporters.swift
```diff
@@ -0,0 +1,56 @@
...
+            at: subdirectoryURL,
+            withIntermediateDirectories: true,
+            attributes: nil
+        )
+    } catch {
+        return nil
+    }
+
+    return subdirectoryURL
+}
```
MustafaHaddara, line 26: This is lifted verbatim from the OTel `Directory` class: https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Exporters/Persistence/Storage/Directory.swift#L51-L65. 

tl;dr we have to create a subdirectory for ourselves before setting up our PersistenceExporters, and if we don't, the **PersistenceExporters fail silently and swallow all of our traces** 😭 

The nitty gritty: we go through 
(a) call `PersistenceMetricExporterDecorator(metricExporter: MetricExporter, storageURL: URL)`
(b) that creates a `PersistenceExporterDecorator<T>(decoratedExporter: T, storageURL: URL)` ([source](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Exporters/Persistence/PersistenceMetricExporterDecorator.swift#L34-L36))
(c) that ends up calling `Directory(url: URL)` -.- ([source](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Exporters/Persistence/PersistenceExporterDecorator.swift#L61))

`Directory` offers a `Directory(withSubdirectoryPath path: String)` constructor but we can't get to that! The main difference in `Directory` constructors is that the one that accepts a `String` will attempt to create the supplied path, if it does not exist. The constructor that accepts a `URL` does no such thing, and assumes that directory already exists. So we have to do that ourselves. 

Options:
1. implement this as-is
2. open an issue against upstream OTel and see if we can get either
    1. create a `PersistenceMetricExporterDecorator ` constructor that takes a path `String` and passes that through
    2. update the behaviour of `Directory` to attempt to create the directory given in a `URL` on init, like it does with a String.

beekhc: I think this is small and straightforward enough we should just keep this implementation.

However, I do think we should either:
* change this to `throws` and return proper errors instead of nil, or at least
* print something before returning nil, so that if it's broken it's not a complete mystery.

Sources/Honeycomb/PersistenceExporters.swift
```diff
@@ -0,0 +1,56 @@
...
+var spanSubdirectoryURL = createCachesSubdirectory("honeycomb/span-cache")!
```
beekhc, line 27: 1. I am a little worried about this being done at static initialization time, since it may involve file system access, and may slow app startup time. Maybe it would be better to call this right before creating the exporter?
2. I am _very_ worried about the `!`. I think if creating the directory fails, we should just skip using the persistence exporter, rather than crashing.

If we move this initialization inside of the functions, then we can catch the error and log it there.

Sources/Honeycomb/PersistenceExporters.swift
```diff
@@ -0,0 +1,56 @@
...
+var metricSubdirectoryURL = createCachesSubdirectory("honeycomb/metric-cache")!
+
+func createPersistenceMetricExporter(_ metricExporter: MetricExporter) -> MetricExporter {
+    do {
+        return try PersistenceMetricExporterDecorator(
+            metricExporter: metricExporter,
+            storageURL: metricSubdirectoryURL
+        )
+    } catch {
+        print(
```
beekhc, line 37: Should we send some telemetry when this happens?

beekhc (commented):
I'm not _too_ worried about the decorator being experimental, but I do think it would be prudent to have a `HoneycombOption` for whether to do offline caching. I think we should have that option in Android, too.

Whether that option defaults to true or false depends on how stable we think the exporter is.

Sources/Honeycomb/PersistenceExporters.swift
MustafaHaddara: I wish, but how? Our telemetry isn't initialized yet 😅 

I suppose I could return an array of errors to report once it has successfully initialized, but that might get messy.

Sources/Honeycomb/Honeycomb.swift
```diff
@@ -96,13 +96,17 @@ public class Honeycomb {
             )
         }
 
-        let spanExporter =
+        var spanExporter =
```
arriIsHere, line 99: Nit: for the sake of consistency we might want to just assign the result of an if statement to the variable similar to how we are selecting protocols.

Sources/Honeycomb/PersistenceExporters.swift
beekhc: Yeah, I felt like I should ask, but I don't really see how we would do it either.

beekhc (approved):

PR #51, opened 2025-03-12
maint: Update to latest opentelemetry-swift

martin308:
Updates to latest `opentelemetry-swift` version

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,7 +16,7 @@ let package = Package(
         .library(name: "Honeycomb", type: .static, targets: ["Honeycomb"])
     ],
     dependencies: [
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -6,7 +6,7 @@ import PackageDescription
 let package = Package(
     name: "honeycomb-opentelemetry-swift",
     platforms: [
-        .macOS(.v10_15),
+        .macOS(.v12),
```
MustafaHaddara, line 9: did this get bumped upstream?

MustafaHaddara (approved):
:+1:

Package.swift
martin308: it did yeh, we tidied up the swift package version files

PR #54, opened 2025-03-12
feat: Use the default resources

martin308:
There is a set of default resources that gets application, device and OS resource information for us. 

Attributes they populate:
- service.name
- service.version
- device.model.identifier
- device.id
- os.type
- os.description
- os.name
- os.version

### DIFF
--- a/smoke-tests/smoke-e2e.bats
+++ b/smoke-tests/smoke-e2e.bats
@@ -196,7 +196,7 @@ mk_diag_attr() {
   result=$(span_names_for "io.honeycomb.view" | sort | uniq -c)
   assert_equal "$result" '   7 "View Body"
    7 "View Render"'
-   
+
   # the View Render spans are tracking the views we expect
   total_duration=$(attribute_for_span_key "io.honeycomb.view" "View Render" "view.name" string | sort)
   assert_equal "$total_duration" '"expensive text 1"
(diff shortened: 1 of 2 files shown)


beekhc (changes requested):
1. Please add a smoke test to check whether the new resources are actually being set. (We fixed a previous similar issue, so there should hopefully already be similar tests.)
2. You mentioned `service.name`, but we already have logic for adding `service.name`, and I think Honeycomb generally wants developers to supply that, so let's make sure there are good tests for that. https://github.com/honeycombio/honeycomb-opentelemetry-swift/blob/27d4f53ba40e03f091246a6ad330f9bbc63600fe/Sources/Honeycomb/HoneycombOptions.swift#L489

martin308:
@beekhc added smoke tests, we could make more of them `assert_not_empty` instead of `assert_equal` if they will be flaky

martin308:
it looks like the way we are merging the resources any `service.name` defined by a customer will be used

smoke-tests/smoke-e2e.bats
```diff
@@ -25,11 +25,16 @@ teardown_file() {
   assert_not_empty "$result"
 }
 
-@test "SDK had default resources" {
-  result=$(resource_attributes_received \
-      | jq 'select (.key == "telemetry.sdk.language").value.stringValue' \
-      | sort | uniq)
-  assert_equal "$result" '"swift"'
+@test "SDK has default resources" {
+  assert_equal "$(resource_attributes_received | jq 'select (.key == "telemetry.sdk.language").value.stringValue' | uniq)" '"swift"'
```
beekhc, line 29: there's enough redundancy in these assertions, you might consider adding a helper

beekhc (approved):
Sounds good.

PR #56, opened 2025-03-13
docs: trace propagation

arriIsHere:

Users will need to know what steps to take to propagate traces to their own backend. This adds documentation on how to do that.

- Closes #<enter issue here>

You need to create an active span in order for all events to be on the same trace (frontend and backend) this differs
from android where you can just use the Otel Context. Swift uses the span context to set this up.

---

- [X] CHANGELOG is updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,6 @@ Below is an example of adding the headers to a network request.
 import OpenTelemetryApi
 
 private struct HttpTextMapSetter: Setter {
-    
     func set(carrier: inout [String: String], key: String, value: String) {
         carrier[key] = value
     }
@@ -186,7 +185,7 @@ private struct HttpTextMapSetter: Setter {
 
 private let textMapSetter = HttpTextMapSetter()
 
-func makeBackendRequest(data: Data) {
+func makeBackendRequest(data: Data) async throws {
     let url = URL(string: "https://mybackendservice")
     var request = URLRequest(url: url!)
     request.httpMethod = "POST"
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -163,6 +163,71 @@ For `.sessionStarted`:
...
+span. The span's context will be used to generate the headers needed for trace propagation.
+
+Below is an example of adding the headers to a network request.
+
+```swift
+
+import OpenTelemetryApi
+
+private struct HttpTextMapSetter: Setter {
+
```
beekhc, line 181: nit: I'd get rid of this one empty line

README.md
```diff
@@ -163,6 +163,71 @@ For `.sessionStarted`:
...
+private let textMapSetter = HttpTextMapSetter()
+
+func makeBackendRequest(data: Data) {
+    let url = URL(string: "https://mybackendservice")
+    var request = URLRequest(url: url!)
+    request.httpMethod = "POST"
+    request.httpBody = data
+
+    let allHeaders: [String: String] = []
```
beekhc, line 195: this is redundant with line 210?

README.md
```diff
@@ -163,6 +163,71 @@ For `.sessionStarted`:
...
+        setter: textMapSetter
+    )
+
+    allHeaders.forEach({ (key: String, value: String) in
+        request.setValue(value, forHTTPHeaderField: key)
+    })
+
+    let session = URLSession(configuration: URLSessionConfiguration.default)
+
+     let (data, response) = try await session.data(for: request)
```
beekhc, line 225: I think the function needs to be marked as `throws` for this `try` to be valid?

beekhc, line 225: This line should be indented the same as the previous line.

beekhc (approved):
Keeping code samples in docs consistent with actual working code is always a pain, but I think having this in the README is better than not having it.

Approved!

PR #57, opened 2025-03-14
feat: add network connection attributes

beekhc:

It would be nice to have attributes on spans about the type of network connection.

This PR uses the `NetworkStatus` package from OTel to add network-specific attributes. OTel uses this package only in their `URLSession` instrumentation, and they spin up a new `NetworkMonitor` on every request.

However, in this PR, we create one `NetworkMonitor` when the Honeycomb SDK is initialized, and we use a new custom span processor to add the attributes to every span.

I added a new smoke test to verify the attributes. Unfortunately, with the smoke tests, we can only verify the "wifi" value.

---

- [X] CHANGELOG is updated

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -24,6 +24,8 @@ let package = Package(
         .target(
             name: "Honeycomb",
             dependencies: [
+                .product(name: "BaggagePropagationProcessor", package: "opentelemetry-swift"),
+                .product(name: "NetworkStatus", package: "opentelemetry-swift"),
                 .product(name: "OpenTelemetryApi", package: "opentelemetry-swift"),
                 .product(name: "OpenTelemetrySdk", package: "opentelemetry-swift"),
                 .product(name: "OpenTelemetryProtocolExporter", package: "opentelemetry-swift"),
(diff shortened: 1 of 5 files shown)


Package.swift
```diff
@@ -24,14 +24,15 @@ let package = Package(
         .target(
             name: "Honeycomb",
             dependencies: [
+                .product(name: "BaggagePropagationProcessor", package: "opentelemetry-swift"),
+                .product(name: "NetworkStatus", package: "opentelemetry-swift"),
```
arriIsHere, line 28: This will break our cocoapods package. We either need these available upstream or we will need to conditionally disable it like we do with GRPC.

Or we could discuss if its worth it to keep maintaining two package ecosystems.

arriIsHere (changes requested):
As it stands these current changes would break our cocoapods build before we can get a release out.

arriIsHere (approved):

Package.swift
beekhc: https://github.com/honeycombio/honeycomb-opentelemetry-swift/pull/59

But given that Cocoapods is already broken from the `PersistenceExporter`, I don't think that PR should block this one.

PR #59, opened 2025-03-14
maint: remove cocoapods support

beekhc:

With our new dependencies on the `PersistenceExporter` and `NetworkStatus` packages in upstream OTel, our Cocoapods support is broken. Before we can reasonably support and maintain Cocoapods in our repo, we need to do the upstream work to make every OTel library we might use support Cocoapods, and all of those pods need tests to guarantee we don't break them with future new dependencies. So, for now, we are dropping Cocoapods support.

This simply removes the cocoapods file.

The only other cocoapods-related change was guarding the compilation of our GRPC code on whether GRPC is available. I think it's fine to leave that in.

N/A

---

- [*] CHANGELOG is updated


arriIsHere (commented):
We should also remove the conditional disabling of grpc

beekhc:
> We should also remove the conditional disabling of grpc

Should we though? I kinda feel like we should leave that for now. I just don't think it's worth the effort to pick it out, considering we don't particularly recommend GRPC anyway. But I'm happy to discuss it Monday.

arriIsHere (approved):

beekhc:
> > We should also remove the conditional disabling of grpc
> 
> Should we though? I kinda feel like we should leave that for now. I just don't think it's worth the effort to pick it out, considering we don't particularly recommend GRPC anyway. But I'm happy to discuss it Monday.

Okay, I tried making it an optional dependency as discussed offline, but that didn't actually work as expected when I tried it in the SmokeTest app. So, I'm just gonna do what you suggest and remove the conditional disabling. We can always revisit later if it becomes important. Thank you for pushing back and keeping me honest!

PR #61, opened 2025-03-17
fix: sometimes crashes dont flush

arriIsHere:

Due to the flakiness of the underlying `flush` methods, sometimes exception logs do not get sent before the app crashes.

- Closes #<enter issue here>

We are adding a 3 second wait to allow the crash exception to finish exporting.

---

- [X] CHANGELOG is updated

### DIFF
--- a/Sources/Honeycomb/HoneycombUncaughtExceptions.swift
+++ b/Sources/Honeycomb/HoneycombUncaughtExceptions.swift
@@ -1,7 +1,9 @@
 import Foundation
+import OpenTelemetrySdk
 
 internal class HoneycombUncaughtExceptionHandler {
     private static var initialUncaughtExceptionHandler: ((NSException) -> Void)? = nil
+    private static var logProcessor: LogRecordProcessor? = nil
 
     public static func initializeUnhandledExceptionInstrumentation() {
         HoneycombUncaughtExceptionHandler.initialUncaughtExceptionHandler =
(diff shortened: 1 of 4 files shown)


arriIsHere:
Draft till we get the release out to not move the changelog cheese

Sources/Honeycomb/HoneycombUncaughtExceptions.swift
```diff
@@ -1,7 +1,9 @@
 import Foundation
+import OpenTelemetrySdk
 
 internal class HoneycombUncaughtExceptionHandler {
     private static var initialUncaughtExceptionHandler: ((NSException) -> Void)? = nil
+    private static var logProcessor: LogRecordProcessor? = nil
```
beekhc, line 6: Are this and line 2 remnants of a previous experiment? I don't see how they're used.

Sources/Honeycomb/HoneycombUncaughtExceptions.swift
```diff
@@ -10,6 +12,9 @@ internal class HoneycombUncaughtExceptionHandler {
         NSSetUncaughtExceptionHandler { exception in
             Honeycomb.log(exception: exception, thread: Thread.current)
 
+            // Wait
```
beekhc, line 15: I'd probably elaborate in the comment on why we need to wait here, so that we'll know when it's safe to remove it later.

beekhc (changes requested):

Sources/Honeycomb/HoneycombUncaughtExceptions.swift
arriIsHere: Yes (oop!)

beekhc (approved):

PR #62, opened 2025-03-17
docs: update RELEASING.md

MustafaHaddara:
I ran into some stuff that wasn't documented when I did the most recent release. 

---
~~- [ ] CHANGELOG is updated~~ N/A
~~- [ ] README is updated with documentation~~ N/A

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -4,5 +4,8 @@
 - Update `CHANGELOG.md` with the changes since the last release.
 - Commit changes, push, and open a release preparation pull request for review.
 - Once the pull request is merged, fetch the updated `main` branch.
-- Apply a tag for the new version on the merged commit (e.g. `git tag -a v1.2.3 -m "v1.2.3"`)
-- Push the tag upstream (this will kick off the release pipeline in CI) e.g. `git push origin v1.2.3`
+- Apply a tag for the new version on the merged commit (e.g. `git tag -a 1.2.3 -m "1.2.3"`) The tag **must** be just the version number (ie. should not start with a `v`) in order for Xcode to pick it up
+- Push the tag upstream (this will kick off the release pipeline in CI) e.g. `git push origin 1.2.3`
+- Copy `CHANGELOG.md` entries for newest version into draft GitHub release created as part of CI publish steps.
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -4,5 +4,8 @@
 - Update `CHANGELOG.md` with the changes since the last release.
 - Commit changes, push, and open a release preparation pull request for review.
 - Once the pull request is merged, fetch the updated `main` branch.
-- Apply a tag for the new version on the merged commit (e.g. `git tag -a v1.2.3 -m "v1.2.3"`)
-- Push the tag upstream (this will kick off the release pipeline in CI) e.g. `git push origin v1.2.3`
+- Apply a tag for the new version on the merged commit (e.g. `git tag -a 1.2.3 -m "1.2.3"`) The tag **must** be just the version number (ie. should not start with a `v`) in order for Xcode to pick it up
+- Push the tag upstream (this will kick off the release pipeline in CI) e.g. `git push origin 1.2.3`
```
beekhc, line 8: "this will kick off the release pipeline in CI"

there is no release pipeline in CI for swift, right?

RELEASING.md
```diff
@@ -4,5 +4,8 @@
...
+- Copy `CHANGELOG.md` entries for newest version into draft GitHub release created as part of CI publish steps.
```
beekhc, line 9: Our latest release had the generated ones, but not stuff copied from CHANGELOG, right? Do we want to go manually update the old ones?

beekhc (changes requested):

RELEASING.md
MustafaHaddara: I generated them and then manually moved them around from section to section.

MustafaHaddara: > Do we want to go manually update the old ones?

I don't think we need to go that far.

beekhc: >> Do we want to go manually update the old ones?
>
> I don't think we need to go that far.

Fair enough.

nit: I'm not sure what "created as part of CI publish steps" means. Should that be deleted as well?

beekhc (approved):

PR #63, opened 2025-03-19
feat: Add UIDevice Attributes

MustafaHaddara:
Add [UIDevice](https://developer.apple.com/documentation/uikit/uidevice) attributes to outgoing spans

Used a span processor instead of resource attributes, because some of these values can change (eg. `batteryLevel`) 

- [x] smoke tests pass

---

- [x] CHANGELOG is updated


martin308 (commented):
~should this be a span processor or a resource detector like the default ones we added recently?~

Just saw the comment in the PR description 🤦

martin308 (commented):
does this compile on macOS? I think UIKit is not available

Sources/Honeycomb/UIDeviceSpanProcessor.swift
```diff
@@ -0,0 +1,112 @@
...
+    public let isStartRequired = true
+    public let isEndRequired = false
+
+    private let device: UIDevice
+    init(_ collectBatteryState: Bool = false) {
+        device = UIDevice.current
+        if (collectBatteryState) {
+            device.isBatteryMonitoringEnabled = true
+        }
+    }
```
MustafaHaddara, line 16: I went back and forth on whether this should be an initialization option or if we should tell clients to set this field themselves...I opted for a config option for consistency with all of the other functionality we have

However, this particular case looks like it's global state, and affects things across the application, so... ¯\\\_(ツ)_/¯

smoke-tests/smoke-e2e.bats
```diff
@@ -143,6 +143,15 @@ mk_attr() {
 
    assert_equal "$result" '"SampleRate"
 "app.metadata"
+"device.isBatteryMonitoringEnabled"
```
MustafaHaddara, line 146: note the lack of `device.batteryLevel` and `device.batteryState` here, that's because the simulator just doesn't have a battery -.-

Sources/Honeycomb/UIDeviceSpanProcessor.swift
beekhc: yeah, i'm a little hesitant to include this one... the fact that it defaults to off for the OS makes me wonder if it's secretly expensive or something. and i'm not sure if it's really all that useful in understanding the app's performance. i think we might want to consider this some more before we check it in

Sources/Honeycomb/HoneycombOptions.swift
```diff
@@ -245,6 +249,8 @@ public struct HoneycombOptions {
 
         private var offlineCachingEnabled: Bool = false
 
+        private var batteryStateAttributesEnabled: Bool = false
```
beekhc, line 252: why should this have its own option, as opposed to just always being included, like most of the other metadata?

beekhc (changes requested):

Sources/Honeycomb/UIDeviceSpanProcessor.swift
MustafaHaddara: Yeah we could totally rip the battery stuff out.

README.md
```diff
@@ -109,6 +109,35 @@ To manually send a span:
...
+    - `device.name` - [UIDevice.name](https://developer.apple.com/documentation/uikit/uidevice/name)
+    - `device.systemName` - [UIDevice.systemName](https://developer.apple.com/documentation/uikit/uidevice/systemname)
+    - `device.systemVersion` - [UIDevice.systemVersion](https://developer.apple.com/documentation/uikit/uidevice/systemversion)
+    - `device.model` - [UIDevice.model](https://developer.apple.com/documentation/uikit/uidevice/model)
+    - `device.localizedModel` - [UIDevice.localizedModel](https://developer.apple.com/documentation/uikit/uidevice/localizedmodel)
+    - `device.userInterfaceIdiom` - [UIDevice.userInterfaceIdiom](https://developer.apple.com/documentation/uikit/uidevice/userinterfaceidiom)
+    - `device.isMultitaskingSupported` - [UIDevice.isMultitaskingSupported](https://developer.apple.com/documentation/uikit/uidevice/ismultitaskingsupported)
+    - `device.orientation` - [UIDevice.orientation](https://developer.apple.com/documentation/uikit/uidevice/orientation)
+    - `device.isLowPowerModeEnabled` - If the user has Low Power Mode enabled on their device.
```
MustafaHaddara, line 136: not technically a `UIDevice` attribute (we get it from `ProcessInfo.processInfo.isLowPowerModeEnabled`) but it felt like it fit here in the README.

Sources/Honeycomb/UIDeviceSpanProcessor.swift
MustafaHaddara: ok we discussed this offline, we came to the conclusion that we won't offer a config-level setting for this but if `UIDevice.current.isBatteryMonitoringEnabled` IS true, we'll include battery info on our spans.

I've made that change now.

martin308 (changes requested):

Sources/Honeycomb/UIDeviceSpanProcessor.swift
```diff
@@ -0,0 +1,111 @@
+import Foundation
+import OpenTelemetryApi
+import OpenTelemetrySdk
+import UIKit
```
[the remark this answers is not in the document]
MustafaHaddara: should be fixed now, it compiles on my mac

martin308 (approved):

beekhc (approved):
This is great! 🚀

PR #64, opened 2025-03-21
feat: navigation instrumentation can track time spent on a screen

MustafaHaddara:
We would like to able to track total time spent in a screen/view/activity

Instead of emitting a single `Navigation` span, we now emit `NavigationTo` and `NavigationFrom` spans. `NavigationTo` spans are the equivalent of the old `Navigation` spans, and the `NavigationFrom` span is entirely new. It includes attributes for the screen we are leaving, as well as time spent on that screen. 

We also hook into UIKit app lifecycle to ensure that the timing of those spans is correct (ie. we're not counting time spent while the phone was locked or the app was in the background). 

Navigation spans now include a `navigation.trigger` attribute, which is either `navigation` or an application lifecycle method (ex. `appDidEnterBackground` etc.). This attribute can also be set by clients via an optional `reason: String` attribute on all of our navigation helpers. 

This PR also fixes the `[]` array issue by explicitly looking for empty paths and emitting `/` instead. We also now accept an optional `prefix: String` param that gets prepended to the path, reducing the likelihood of empty paths and allowing clients to distinguish between different NavigationStacks in their applications. 

- [x] smoke tests pass

---

- [x] CHANGELOG is updated


Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
```diff
@@ -32,10 +34,20 @@ internal class HoneycombNavigationProcessor {
...
             .setAttribute(key: "screen.name", value: path)
-            .startSpan()
-            .end()
+
+        if lastNavigationTime != nil && lastNavigationSpan != nil {
+            let lastScreenNavigationTime = Date().timeIntervalSince(lastNavigationTime!)
+            lastNavigationSpan!
+                .setAttribute(key: "screen.active.time", value: Double(lastScreenNavigationTime))
+
+            // todo: hook into lifecycle to ensure this span always ends?
```
beekhc, line 45: We would need to, but I think it's worse than that. If the app crashes, I don't think there's any way to get notified to end the span, and even if we did end the span during shutdown, it wouldn't have time to flush properly. So we'd be removing the ability to see what screen the app was on during crashes.

I think if we do this, we'd want to do something in _addition_ to the previous implementation.

Finding the right place to hook in for a lifecycle event might be difficult. We'd probably want to swizzle the app's AppDelegate, but that's a medium lift. I do think there's a good chance we'll want to do it eventually anyway, for other Core Mobile Vitals signals.

Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
```diff
@@ -32,10 +34,20 @@ internal class HoneycombNavigationProcessor {
...
+            lastNavigationSpan!.end()
```
beekhc, line 46: I think we should consider what the "right" behavior is for nested screens, since we support that kind of thing. I mean, in the case of tabs, for example, if we're setting the current tab as the screen, it's going to look like the user visited the current underlying screen and quickly navigated away to the specific tab. I guess that's fine. It'll probably be pretty common in Android apps that use fragments. I dunno. I guess it's fine, but just something we should consider.

README.md
```diff
@@ -281,9 +281,15 @@ struct SampleNavigationView: View {
 }
 ```  
 
-Whenever the `path` variable changes, this View Modifier will emit a span with the name `Navigation`. This span will contain the following attributes:
+Whenever the `path` variable changes, this View Modifier will emit a span with the name `NavigationTo`. This span will contain the following attributes:
 
 - `screen.name` (string): the full navigation path when the span is emitted. If the path passed to the view modifier is not `Encodable` (ie. if you're using a `NavigationPath` and have stored a value that does not conform to the `Codable` protocol), then this attribute will have the value `<unencodable path>`.
+- `navigation.trigger`: Normally `navigation`. May also be `appDidBecomeActive` if the app moves into the foreground and still has navigation context.
```
beekhc, line 287: > Normally `navigation`

That's not bad, but I wonder if we could do better -- maybe be more explicit. If `reportNavigation` had an optional parameter to set this, would that be useful? 

Then it would be, like, "Defaults to "navigation", unless you provide a different trigger to `reportNavigation`."

I dunno. Does that sound useful?

MustafaHaddara: yeah definitely! I can add that.

Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
```diff
@@ -16,76 +21,158 @@ func getTracer() -> Tracer {
...
-    func reportNavigation(path: [Encodable]) {
+    func reportNavigation(path: [String], reason: String? = nil) {
+        // If we have a previous navigation, emit a "NavigationFrom" span
+        navigationEnd(reason: reason ?? "navigation")
+
+        // Update current path
+        currentNavigationPath = path
+
+        navigationStart(reason: reason ?? "navigation")
+    }
```
MustafaHaddara, line 63: this change is subtle, might not pop up right away (and we didn't have tests exercising this) 

the navigation split view in the examples builds an explicit path, ex

```swift
let path: [Encodable] = ["Split View", park, tree]
Honeycomb.setCurrentScreen(path: path)
```

and previously this would hit the function definition below (which accepts a `[Encodable]`). That would get stringified into a single long string and get used as both the path _and_ the current screen name 

instead, now we preserve the array that the client gave us as the full path and just use the last entry as the current screen name

Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
```diff
@@ -16,76 +21,177 @@ func getTracer() -> Tracer {
...
 
-            reportNavigation(path: pathStr)
+            reportNavigation(path: pathStr, reason: reason)
         } catch {
-            reportNavigation(path: unencodablePath)
+            reportNavigation(path: unencodablePath, reason: reason)
+        }
+    }
+
+    func reportNavigation(prefix: String? = nil, path: [String], reason: String? = nil) {
```
MustafaHaddara, line 55: it's a little subtle to spot in the git diff but this is an actual functional change here.

Previously, in the version of this function that accepted a `path: [Encodable]`, we would stringify that entire path into a _single_ string. This meant that if that path was empty, our navigation instrumentation would receive `"[]"` (as a string!) as the path. 

Now, that function turns the `[Encodable]` into a `[String]` and calls this function, which calls `navigationStart()`. That function looks at the path and uses the last item in the array as `screen.name`, and uses the `/`-separated version of the array as the `screen.path` attribute.

beekhc (approved):
awesome!

PR #65, opened 2025-03-28
maint: add CI for all the platforms that we support

martin308:

We declare in our `package.swift` file that we support the following platforms:

```
    platforms: [
        .macOS(.v12),
        .iOS(.v13),
        .tvOS(.v13),
        .watchOS(.v6),
    ],
```

But we don't actually run any of our CI for these. We don't actually compile on all of these platforms.

Added CI to at least compile the SDK on each of the platforms that we say we support

Added a bunch of hacky conditional compilation statements so that build passes. It's possible we can structure some of the code better to make these a little less unweildly. This will at least put us on the right path of supporting all the platforms we say we do.

We will probably want to update the readme to specify which platforms support which things

Build passes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,7 @@ jobs:
     steps:
       - checkout
       - run: xcodebuild -version
-      - run: xcodebuild test -scheme honeycomb-opentelemetry-swift -destination << parameters.destination >>
+      - run: xcodebuild -scheme honeycomb-opentelemetry-swift -destination << parameters.destination >>
 
   tests:
     macos:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,18 @@ jobs:
       - checkout
       - run: make lint
 
+  build:
+    parameters:
+      destination:
+        type: string
+    macos:
+      xcode: 15.4.0
+    resource_class: macos.m1.medium.gen1
+    steps:
+      - checkout
+      - run: xcodebuild -version
+      - run: xcodebuild test -scheme honeycomb-opentelemetry-swift -destination << parameters.destination >>
+
   tests:
     macos:
       xcode: 15.4.0
@@ -86,6 +98,14 @@ workflows:
                 - main
     jobs:
       - lint
+      - build:
+          matrix:
+            parameters:
+              destination:
+                - generic/platform=macOS
+                - generic/platform=iOS
+                - generic/platform=tvOS
+                - generic/platform=watchOS
       - tests
       - smoke_tests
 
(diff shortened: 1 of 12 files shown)


.circleci/config.yml
```diff
@@ -19,6 +19,18 @@ jobs:
       - checkout
       - run: make lint
 
+  build:
```
beekhc, line 22: doesn't this conflict with the other job called "build" below?

Sources/Honeycomb/AttributeValueExtensions.swift
```diff
@@ -0,0 +1,54 @@
...
+}
+
+extension [String]: AttributeValueConvertable {
+    func attributeValue() -> AttributeValue {
+        AttributeValue.array(AttributeArray(values: self.map { AttributeValue.string($0) }))
+    }
+}
+
+#if canImport(MetricKit)
+    extension TimeInterval: AttributeValueConvertable {
```
beekhc, line 35: I think TimeInterval is part of Foundation, so this extension can go outside the block?

Sources/Honeycomb/Honeycomb.swift
```diff
@@ -30,8 +33,10 @@ private func createKeyValueList(_ dict: [String: String]) -> [(String, String)]
 }
 
 public class Honeycomb {
-    @available(iOS 13.0, macOS 12.0, *)
-    static private let metricKitSubscriber = MetricKitSubscriber()
+    #if canImport(MXMetricManager)
+        @available(iOS 13.0, macOS 12.0, *)
```
beekhc, line 37: isn't the `@available` redundant with the `canImport`? it's unclear to me why we'd need both.

same for the place where it's set below.

.circleci/config.yml
martin308: there is a build workflow but I dont see another job named build. Do you have a line number?

Sources/Honeycomb/AttributeValueExtensions.swift
martin308: will see if it works!

martin308: TimeInterval is not available on watchOS

martin308: I was wrong!

martin308: managed to remove all of the conditionals from this file 🎉

.circleci/config.yml
beekhc: oh, duh, github had collapsed the header where it switched from jobs to workflows, and i didn't notice. never mind!

beekhc (approved):

PR #67, opened 2025-04-07
fix: navigation instrumentation no longer serializes path as empty array

MustafaHaddara:
- Navigation instrumentation was still reporting paths as `[]` when clients were using a `NavigationPath` to store their path (instead of an array)

Change how we encode `NavigationPath` objects for our instrumentation.

- [x] smoke tests pass

---

- [x] CHANGELOG is updated
- [x] ~~README is updated with documentation~~ N/A

### DIFF
--- a/Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
+++ b/Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
@@ -30,7 +30,15 @@ internal class HoneycombNavigationProcessor {
     @available(tvOS 16.0, iOS 16.0, macOS 13.0, watchOS 9, *)
     func reportNavigation(prefix: String? = nil, path: NavigationPath, reason: String? = nil) {
         if let codablePath = path.codable {
-            reportNavigation(path: codablePath, reason: reason)
+            // Janky hack: round trip the NavigationPath to get the individual elements in an array
+            // NavigationPath doesn't offer a API for this
+            // it also (from our testing) seems like it tends to encode each step of the path as two items:
+            // the class name and then the actual object
+            // but since this is an entirely undocumented internal structure, we're not going to make any assumptions there
+            // and will just encode the full path as we get it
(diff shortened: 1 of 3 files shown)


Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
```diff
@@ -30,14 +30,22 @@ internal class HoneycombNavigationProcessor {
...
         if let codablePath = path.codable {
-            reportNavigation(path: codablePath, reason: reason)
+            // Janky hack: round trip the NavigationPath to get the individual elements in an array
+            // NavigationPath doesn't offer a API for this
+            // it also (from our testing) seems like it tends to encode each step of the path as two items:
+            // the class name and then the actual object
+            // but since this is an entirely undocumented internal structure, we're not going to make any assumptions there
+            // and will just encode the full path as we get it
+            let encodedPath = try! JSONEncoder().encode(codablePath)
+            let decodedPath = try! JSONDecoder().decode([String].self, from: encodedPath)
```
beekhc, line 40: i've gotta think some on this "janky hack", but my immediate feedback is we shouldn't use `try!`, because it would suck if we caused the app to crash.

MustafaHaddara: Makes sense. I was thinking it's ok because that this point we're guaranteed that `codablePath` is, well, codable...and anything encoded by a `JSONEncoder` is also guaranteed to be decodable by a `JSONDecoder` but I can switch down to just a `try`.

beekhc (approved):

Sources/Honeycomb/HoneycombNavigationSpanInstrumentation.swift
beekhc: I dunno. I'd be worried about the format changing from [String] someday. But anyway, yeah, even if it's something that _should_ never happen, I think it's better for an observability instrumentation to have a sensible fallback than crash. The new implementation looks good.

PR #77, opened 2025-05-07
feat: include stack trace info in MXCrashDiagnostic and MXHangDiagnostic payloads

MustafaHaddara:
Includes stack trace info in MXCrashDiagnostic and MXHangDiagnostic payloads

Both `MXCrashDiagnostic` and `MXHangDiagnostic` objects include a `callStackTree` property of type `MXCallStackTree`. That class has only one useful function: [`jsonRepresenation()`](https://developer.apple.com/documentation/metrickit/mxcallstacktree/jsonrepresentation()). 

So we'll pass that through on the trace and symbolicate it properly in the collector plugin. 

- [x] smoke tests pass

---

- [x] CHANGELOG is updated
- [ ] ~README is updated with documentation~ N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@ Honeycomb OpenTelemetry SDK Changelog
 
 ## v.Next
 
+### New Features
+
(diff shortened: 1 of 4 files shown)


arriIsHere (approved):
💥 🚀

PR #80, opened 2025-05-09
feat: Bring back cocoapods!

beekhc:

This PR brings back CocoaPods support.

1. Bring back the cocoapods spec from before.
2. Add the new dependencies we've added since it was created.
3. Add a test project that uses the cocoapod spec.
4. Add the test project to circleci.
5. Update all tests to use m4pro, which changes runtime from >20min to <10min.

---

- [X] CHANGELOG is updated

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ jobs:
 
     steps:
       - macos/preboot-simulator:
-          version: "18.2"
+          version: "18.5"
           platform: "iOS"
           device: "iPhone 16"
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,7 +47,7 @@ jobs:
 
   cocoapods_tests:
     macos:
-      xcode: 16.2.0
+      xcode: 16.4.0
     resource_class: macos.m1.medium.gen1
 
     steps:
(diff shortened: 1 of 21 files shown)


.circleci/config.yml
```diff
@@ -34,7 +34,7 @@ jobs:
   tests:
     macos:
       xcode: 15.4.0
-    resource_class: macos.m1.medium.gen1
+    resource_class: m4pro.medium
```
MustafaHaddara, line 37: why not m4max? is that not available?

honeycomb-opentelemetry-swift.podspec
```diff
@@ -0,0 +1,42 @@
...
+
+  s.swift_version = "5.10"
+  s.ios.deployment_target = "13.0"
+  s.tvos.deployment_target = "13.0"
+  s.watchos.deployment_target = "6.0"
+
+  s.source_files = 'Sources/Honeycomb/**/*.swift'
+
+  s.dependency 'OpenTelemetry-Swift-Api', '~> 1.17.1'
+  s.dependency 'OpenTelemetry-Swift-Sdk', '~> 1.17.1'
```
MustafaHaddara, line 29: should we update the dependencies in our `Package.swift`? In there we have

```
dependencies: [
    .package(url: "https://github.com/open-telemetry/opentelemetry-swift.git", from: "1.15.0")
],
```

MustafaHaddara (approved):
🎉

.circleci/config.yml
beekhc: nope

https://circleci.com/docs/using-macos/

PR #81, opened 2025-05-21
fix: refine attributes for manual error logging

beekhc:

This refines the attributes for manual error logging to better match semantic conventions, and also makes it possible for us to differentiate fatal _crashes_ from non-fatal _errors_.

* Add optional `severity` parameter to manual error-logging APIs.
* Change logging of `NSError` and Swift `Error` to match semantic conventions for _errors_, not _exceptions_.

Smoke tests are updated.

---

- [X] CHANGELOG is updated

### DIFF
--- a/Sources/Honeycomb/Honeycomb.swift
+++ b/Sources/Honeycomb/Honeycomb.swift
@@ -306,9 +306,10 @@ public class Honeycomb {
 
         // TODO: Type and name seem wrong here. Which is right?
         var errorAttributes = [
-            "exception.type": type.attributeValue(),
-            "exception.message": message.attributeValue(),
-            "exception.stacktrace": exception.callStackSymbols.joined(separator: "\n")
+            SemanticAttributes.exceptionType.rawValue: type.attributeValue(),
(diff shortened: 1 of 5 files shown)


Sources/Honeycomb/Honeycomb.swift
```diff
@@ -262,47 +262,52 @@ public class Honeycomb {
...
-        let type = String(describing: Mirror(reflecting: exception).subjectType)
+        let type = exception.name.rawValue
         let message = exception.reason ?? exception.name.rawValue
 
+        // TODO: Type and name seem wrong here. Which is right?
         var errorAttributes = [
             "exception.type": type.attributeValue(),
             "exception.message": message.attributeValue(),
-            "exception.name": exception.name.rawValue.attributeValue(),
             "exception.stacktrace": exception.callStackSymbols.joined(separator: "\n")
```
martin308, line 311: would be good to reference the semantic conventions from the upstream SDK.

https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/OpenTelemetryApi/Trace/SemanticAttributes.swift#L283

martin308 (approved):

PR #85, opened 2025-06-13
feat: add build info to resource attributes

MustafaHaddara:
Adds resource attributes to enable us to symbolicate stacktraces from our exception logger

The important ones are `app.debug.buildUUID` and `app.debug.binaryName`. In my tests, `app.debug.binaryName` spat out `/Library/Developer/CoreSimulator/Volumes/iOS_22C150/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS 18.2.simruntime/Contents/Resources/RuntimeRoot/usr/lib/libLogRedirect.dylib`, not the actual executable name like I expected. This might be because I was running the app in the xcode simulator; we'll have to experiment with an actual compiled app on a device. 

Anyways I added `app.bundle.executable` to make up for that, which looks like it contains the actual binary name in it.  

While I was in here, I added code to pull the version number info out of the bundle as well, since that was fairly straightforward. 

- [x] Tests pass

---

- [x] CHANGELOG is updated

### DIFF
--- a/Sources/Honeycomb/AppInfoResources.swift
+++ b/Sources/Honeycomb/AppInfoResources.swift
@@ -3,21 +3,27 @@ import MachO
 import OpenTelemetryApi
 import OpenTelemetrySdk
 
+// none of these are in the semconv yet so we can call them whatever we like
+let APP_BUNDLE_VERSION = "app.bundle.version"
+let APP_BUNDLE_SHORT_VERSION_STRING = "app.bundle.shortVersionString"
+let APP_DEBUG_BUILD_UUID = "app.debug.buildUUID"
+let APP_DEBUG_BINARY_NAME = "app.debug.binaryName"
+let APP_BUNDLE_EXECUTABLE = "app.bundle.executable"
+
 public func getAppResources() -> [String: String] {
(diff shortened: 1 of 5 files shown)


Sources/Honeycomb/AppInfoResources.swift
```diff
@@ -0,0 +1,97 @@
...
+import MachO
+import OpenTelemetryApi
+import OpenTelemetrySdk
+
+// none of these are in the semconv yet so we can call them whatever we like
+let APP_BUNDLE_VERSION = "app.bundle.version"
+let APP_BUNDLE_SHORT_VERSION_STRING = "app.bundle.shortVersionString"
+let APP_DEBUG_BUILD_UUID = "app.debug.buildUUID"
+let APP_DEBUG_BINARY_NAME = "app.debug.binaryName"
+let APP_BUNDLE_EXECUTABLE = "app.bundle.executable"
```
MustafaHaddara, line 11: perhaps these should live under the `service.*` namespace instead? ¯\\\_(ツ)_/¯

martin308 (approved):

PR #88, opened 2025-06-25
feat: add mappings for mach exceptions and unix symbols

MustafaHaddara:
Gives us an exception type and message for metrickit crashes

Apple's docs for [`exceptionType`](https://developer.apple.com/documentation/metrickit/mxcrashdiagnostic/exceptiontype) and [`signal`](https://developer.apple.com/documentation/metrickit/mxcrashdiagnostic/signal) pointed me at some header files in the SDK that had these mappings, so I grabbed them from there.

On the collector processor side, we'll need to decide how to grab these and which takes priority. Looking at what we have already in this subscriber, I'm thinking the priority order would be something like:

| order | exception type | exception message |
| - | - | - |
| 1 | `exception.objc.type` | `exception.objc.message` |
| 2 | `exception.mach_exception.name` | `exception.mach_exception.description` |
| 3 | `exception.signal.name` | `exception.signal.description` |
| 4 | "Unknown Error" | `exception.termination_reason` |
| 5 | "Unknown Error" | "Unknown Error" |

- [x] tests pass

---

- [x] CHANGELOG is updated
- [x] ~README is updated with documentation~ N/A


Sources/Honeycomb/MetricKitSubscriber.swift
```diff
@@ -419,10 +419,22 @@
                 attrs["exception.code"] = exceptionCode.intValue
             }
             if let exceptionType = $0.exceptionType {
-                attrs["exception.mach_execution_type"] = exceptionType.intValue
```
MustafaHaddara, line 422: note this used to be spelled `execution_type` instead of `exception_type`, I assumed that was an error

martin308 (approved):
I think this looks right

PR #89, opened 2025-06-26
rel: prep for 0.0.14 release

MustafaHaddara:


martin308 (approved):
one header to add back to the release notes then it looks good

PR #91, opened 2025-06-26
maint: Upload collector output as an artifact

martin308:

Upload the collector output as an artifact so that it can be downloaded and inspected.

see the json file as an artifact


arriIsHere (approved):
Looks like circle is adding it to the artifacts so LGTM
🚀

PR #93, opened 2025-07-24
docs: make readme consistent with android

arriIsHere:
We had some insconsistencies between our IOS and android readmes.

- Closes #<enter issue here>

Update the docs in various sections to make the IOS and android Readmes consistent.

---

- [ ] CHANGELOG is updated


jairo-mendoza (approved):
lgtm! ty!

PR #97, opened 2025-08-12
feat: make HoneycombOptions builder object ObjectiveC friendly

arriIsHere:

To support react native we need to allow access to the options builder to ObjectiveC.

- Closes #<enter issue here>

Make builder inherit `NSObject` and give it and applicable methods the `@objc` directive.

---

- [X] CHANGELOG is updated

### DIFF
--- a/Sources/Honeycomb/HoneycombOptions.swift
+++ b/Sources/Honeycomb/HoneycombOptions.swift
@@ -216,7 +216,7 @@ public struct HoneycombOptions {
 
     let offlineCachingEnabled: Bool
 
-    @objc public class Builder : NSObject {
+    @objc public class Builder: NSObject {
         private var apiKey: String? = nil
         private var tracesApiKey: String? = nil
         private var metricsApiKey: String? = nil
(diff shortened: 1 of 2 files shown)


Sources/Honeycomb/HoneycombOptions.swift
```diff
@@ -471,33 +471,33 @@ public struct HoneycombOptions {
             return self
```
arriIsHere, line 471: These methods cannot be set as @objc functions due to the argument type not being compatible.

arriIsHere:
the `build` method is not being included as the `HoneycombOptions` object is not capable of being converted into an ObjectiveC compatible class.

Sources/Honeycomb/HoneycombOptions.swift
beekhc: Totally makes sense.

beekhc (approved):

beekhc:
> the `build` method is not being included as the `HoneycombOptions` object is not capable of being converted into an ObjectiveC compatible class.

is that because it's a `struct`? ugh. yeah, i guess it's not worth changing. i guess it's probably better to let our `configure` method take the builder in objc than to make the `struct` a `class` in swift.

PR #105, opened 2025-09-05
maint: remove cocoapods again

beekhc:

* We want to update to OpenTelemetry 2.0.
* OpenTelemetry Swift's Cocoapods release has been broken since 2.0.
* Cocoapods is already being phased out in favor of Swift Package Manager, and set to be made readonly next year.
* Cocoapods is no longer necessary for React Native support.

Therefore, we are dropping Cocoapods support.

Deletes the cocoapods config, and also the CI tests that verified it was working.

N/A

---

- [X] CHANGELOG is updated


arriIsHere (approved):
🟥 ➖ 🎉

PR #106, opened 2025-09-10
feat: expose OpenTelemetry Resource as public property

wolfgangcodes:

Adds a public `resource` property to the `Honeycomb` class that exposes the configured OpenTelemetry Resource. This matches the Android SDK functionality and provides developers with access to inspect resource attributes for debugging and validation purposes.

**Key changes:**
- Added `public static var resource: Resource` property to `Honeycomb` class
- Initializes with default OpenTelemetry Resource
- Updates to fully configured resource after `configure(options:)` is called

1. **After configuration**: `Honeycomb.resource` returns fully configured resource with custom attributes
2. **Access attributes**: `Honeycomb.resource.attributes["service.name"]` provides service name
3. **Tests**: Run `swift test --filter testResourceExposureAfterConfiguration` to verify functionality

---

- [x] CHANGELOG is updated
- [ ] ~~README is updated with documentation~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ Honeycomb OpenTelemetry SDK Changelog
 
 ## v.Next
 
+* feat: expose OpenTelemetry Resource as public property
+
 ## 2.0.0
 
 * maint: Update to OpenTelemetry 2.x.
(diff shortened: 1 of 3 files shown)


arriIsHere (approved):
💌

PR #108, opened 2025-09-21
fix: Update to use otel-swift-core 2.1.1 and otel-swift 2.1.0

irvingpop:
I was getting package resolution errors in Xcode when adding the honeycomb-opentelemetry-swift package, due to the upstream repo splitting up.
```
product ‘OpenTelemetryApi’ required by package ‘honeycomb-opentelemetry-swift’ target ‘Honeycomb’ not found in package ‘opentelemetry-swift’.
```

This PR points all the dependencies in the right places, and lets me successfully add the package in Xcode.
Apologies about the gratuitous whitespace cleanup in the README

Verified manually by adding my fork repo as a package dependency

---

- [ ] CHANGELOG is updated

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,15 +1,6 @@
 {
-  "originHash" : "8bf0b58cdca5a499330b34a827aba40cb4658fbcaac94b40fa8c8dd2208dab6c",
+  "originHash" : "b65e2a5736041bad324c421d028cef43b92e3a01953132fe97fb0838e5af1dad",
   "pins" : [
-    {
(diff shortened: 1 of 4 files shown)


beekhc (commented):
I think this looks good, but before I approve, I want to look into something with our CircleCI tests, which test against specific OTel versions. I would expect this change to break those, and I just want to double-check they are working correctly.

beekhc (approved):
I figured out why the circleci test isn't failing, and it's fine for now.

Thanks for doing the upgrade!

PR #111, opened 2025-10-02
fix: add session ID to log records

arriIsHere:

- Fixes missing session ID on log records

Log records were not including the session ID attribute that spans already had. This PR adds a new `HoneycombSessionIdLogRecordProcessor` that wraps the log processor chain and adds the `session.id` attribute to all log records, ensuring consistency with span data.

- Run the smoke tests - they now verify that all log records include a `session.id` attribute
- Check that logged exceptions and errors include the session ID in Honeycomb

---

- [x] CHANGELOG is updated

### DIFF
--- a/Sources/Honeycomb/Session/HoneycombSessionIdLogRecordProcessor.swift
+++ b/Sources/Honeycomb/Session/HoneycombSessionIdLogRecordProcessor.swift
@@ -16,7 +16,7 @@ struct HoneycombSessionIdLogRecordProcessor: LogRecordProcessor {
 
         enhancedRecord.setAttribute(
             key: SemanticConventions.Session.id,
-            value: AttributeValue.string(sessionManager.sessionId)
+            value: sessionManager.sessionId
         )
 
         nextProcessor.onEmit(logRecord: enhancedRecord)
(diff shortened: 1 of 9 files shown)


Sources/Honeycomb/Session/HoneycombSessionIdLogRecordProcessor.swift
```diff
@@ -0,0 +1,32 @@
...
+    init(nextProcessor: LogRecordProcessor, sessionManager: HoneycombSessionManager) {
+        self.nextProcessor = nextProcessor
+        self.sessionManager = sessionManager
+    }
+
+    public func onEmit(logRecord: ReadableLogRecord) {
+        var enhancedRecord = logRecord
+
+        enhancedRecord.setAttribute(
+            key: "session.id",
```
wolfgangcodes, line 18: Does swift have a semantic constant we can rely on?

arriIsHere: I'll see if we can update that in the span processor as well.

wolfgangcodes (approved):

PR #113, opened 2025-10-10
fix: make HoneycombNavigationPathSpanProcessor threadsafe

beekhc:

We've had reports of crashes in the `HoneycombNavigationPathSpanProcessor`, and it looks like the handling of the current path was not threadsafe, so this PR makes it threadsafe.

We're just guarding each access of the current path members with an `NSLock`, similar to what we did in the Session Manager.

The smoke tests still pass.

---

- [X] CHANGELOG is updated
~- [ ] README is updated with documentation~ N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,17 +2,19 @@ Honeycomb OpenTelemetry SDK Changelog
 
 ## v.Next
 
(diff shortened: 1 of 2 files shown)


arriIsHere (approved):
🧵 🔒

MustafaHaddara (approved):
👍

PR #119, opened 2025-11-12
fix: avoid MetricKit crash by initializing CTTelephonyNetworkInfo on main thread

### LINKED ISSUE #118: MetricKit diagnostic callback triggers crash in NetworkStatusSpanProcessor
**Versions**

honeycomb-opentelemetry-swift: main branch / 2.2.0

iOS 17.x

MetricKit enabled

**Steps to reproduce**

Use the Honeycomb OpenTelemetry Swift SDK with a NetworkStatusSpanProcessor configured.

Allow MetricKit to deliver diagnostic payloads (eg, via MXMetricManager).

When MetricKitSubscriber.didReceive runs, it creates a span to report diagnostics.

During that span creation, NetworkStatusSpanProcessor.onStart is called.

The processor builds a new NetworkStatus, which internally initializes CTTelephonyNetworkInfo.

The synchronous XPC setup that happens inside CTTelephonyNetworkInfo.init() can hang or deadlock when called from a background queue like com.apple.metrickit.manager.queue.

**Expected Behaviour**

CTTelephonyNetworkInfo should be initialized safely on the main thread, or reused if already created.

Span creation should be lightweight and safe on all threads, including MetricKit’s background queue.

**Observed Behaviour**

Occasional crashes when MetricKit delivers diagnostics, with stack traces pointing to CTTelephonyNetworkInfo being initialized inside NetworkStatusSpanProcessor.onStart().

The issue appears to stem from synchronous XPC communication during telephony client setup on a background queue.

**Proposed Fix**

Move CTTelephonyNetworkInfo initialization to the main thread (either at SDK setup time or via a main-thread hop).

Set up and keep a single NetworkStatus instance instead of creating a new one for every span. Since NetworkStatus holds a reference to a NetworkMonitor rather than fixed values, it should still pick up the current network state when used.

Compare: https://github.com/honeycombio/honeycomb-opentelemetry-swift/compare/main...andy-dufour:honeycomb-opentelemetry-swift:adf/cttelephony-thread-fix?expand=1

**Additional context**

The crash was observed through Crashlytics traces showing XPC calls like _connect_sync on the MetricKit manager queue. Unfortunately only happening at scale, intermittent and difficult to reproduce.

Initializing CTTelephonyNetworkInfo on the main thread prevents these hangs and ensures safe XPC communication.

andy-dufour:

Closes #118
Fixes a crash that occurred when CTTelephonyNetworkInfo was initialized from MetricKit’s background queue during diagnostic delivery.

This PR moves CTTelephonyNetworkInfo initialization to the main thread to avoid unsafe synchronous XPC calls that can hang or crash when invoked from background queues like com.apple.metrickit.manager.queue.

NetworkStatusSpanProcessor now:

Sets up and keeps a single NetworkStatus instance instead of creating a new one for every span. Since NetworkStatus holds a reference to a NetworkMonitor rather than fixed values, it should still pick up the current network state when used.

This change aims to avoid the MetricKit crash while keeping network attributes accurate and up to date.

It's an intermittent result, it will need to be validated at scale, but:

Run an app instrumented with this SDK and MetricKit enabled.

Wait for MetricKit diagnostics to trigger

Verify that spans are created successfully and no crash occurs in CTTelephonyNetworkInfo initialization.

To ensure the NetworkMonitor is updating properly after init: Confirm network-related span attributes (e.g., carrier, radio type) continue to update as expected when the device’s network changes.

---

- [ X] CHANGELOG is updated
- [ ] README is updated with documentation - not required for this fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ Honeycomb OpenTelemetry SDK Changelog
 
 ## v.Next
 
+* fix: Fix crash caused by initializing CTTelephonyNetworkInfo on a background thread during MetricKit diagnostics. Initialization now happens safely on the main thread.
+
 ## 2.2.0
 
 * feat: add device.manufacturer and device.model.name attributes
--- a/Sources/Honeycomb/NetworkStatusSpanProcessor.swift
+++ b/Sources/Honeycomb/NetworkStatusSpanProcessor.swift
@@ -19,6 +19,11 @@
             // This ensures CTTelephonyNetworkInfo is initialized during Honeycomb setup
             // (typically on main thread) rather than on background queues like MetricKit's,
             // which can cause hangs and XPC connection issues.
+            //
+            // Note: NetworkStatus holds a reference to NetworkMonitor, which actively
+            // monitors network state changes. When inject() is called, it should query
+            // the monitor for current state rather than using stale cached data.
+            // This provides both thread safety and up-to-date network information.
             networkStatus = NetworkStatus(with: monitor)
             injector = NetworkStatusInjector(netstat: networkStatus)
         }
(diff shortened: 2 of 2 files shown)


MustafaHaddara (approved):
Looks good to me, but let's wait for @beekhc to look at it before we merge.

Thanks!

beekhc (approved):
I am not 100% convinced that this will fix the crashes completely, but I am convinced that this is an improvement, so I approve.

PR #125, opened 2025-12-14
feat: Add options for using OpenTelemetry URLSession and MetricKit instrumentation instead of Honeycomb versions

beekhc:

Adds options to use upstream URLSession and MetricKit instrumentation.

The URLSession instrumentation defaults to true (using the upstream version), because the change should be invisible to our customers. I only left the old instrumentation in there for now in case there is some problem when customers upgrade, so they can switch back without waiting for us to make a new release.

The MetricKit instrumentation defaults to false for now, because there are minor differences between the upstream version and ours. I have [a PR out](https://github.com/open-telemetry/opentelemetry-swift/pull/1014) to iron out those differences. Once they are identical, we can switch this to default to true.

Next time there's a major release, we should remove these options and remove our own versions of this instrumentation.

Smoke tests still pass with the new defaults. I have also tried them with the other values manually and verified they still work (with the expected changes to MetricKit telemetry).

---

- [X] CHANGELOG is updated
N/A ~- [ ] Update `honeycombLibraryVersion` in `Sources/Honeycomb/HoneycombVersion.swift`~


MustafaHaddara (approved):
Works for me.

I left one optional question.

PR #134, opened 2026-07-01
rel: 2.4.1 — fix crash from OTel URLSessionInstrumentation default

jclr:
Fixes AGENTO-378 (https://linear.app/honeycombio/issue/AGENTO-378/app-crash-after-otel-swift-sdk-240-update)

### Summary

Fixes a crash on `Honeycomb.configure` introduced in **2.4.0**.

2.4.0 (#125) added `otelUrlSessionInstrumentationEnabled` and defaulted it to `true`, switching all consumers from Honeycomb's own URLSession instrumentation to OpenTelemetry's `URLSessionInstrumentation`. Its initializer runs `injectInNSURLClasses()`, which uses `DispatchQueue.concurrentPerform` to call `class_copyMethodList` over every ObjC class **in parallel**. That forces concurrent ObjC class realization — which is **not thread-safe** — and segfaults (`EXC_BAD_ACCESS` in `realizeClassWithoutSwift`) during init, especially in apps with many lazily-realized ObjC classes (e.g. Firebase-heavy apps).

This is an **unfixed upstream bug** in opentelemetry-swift (`concurrentPerform` is still present on `main`), so bumping the dependency does not fix it. This PR defaults the option back to `false`, restoring the stable pre-2.4.0 behavior (Honeycomb's own URLSession instrumentation, still enabled by default). The OTel path remains available as an explicit opt-in via `.setOTELURLSessionInstrumentationEnabled(true)`.

### Changes

- `HoneycombOptions.swift`: default `otelUrlSessionInstrumentationEnabled` `true` → `false`
- `HoneycombOptionsTests.swift`: assert new default + cover the setter (previously untested)
- `CHANGELOG.md`: `## 2.4.1` entry
- `HoneycombVersion.swift`: bump to `2.4.1`

### Release checklist
- [x] CHANGELOG is updated
- [x] `honeycombLibraryVersion` bumped in `Sources/Honeycomb/HoneycombVersion.swift`
- [ ] README updated with documentation (n/a — dependency table tracks opentelemetry-swift versions, not the SDK version)

### Follow-up
File an upstream issue against opentelemetry-swift for the non-thread-safe concurrent class realization in `injectInNSURLClasses()` (draft prepared separately; related existing reports: #522, #937, #229).


tanya-coding (commented):
LGTM but would welcome reviews from folks more familiar with the area

MustafaHaddara (approved):

Sources/Honeycomb/HoneycombOptions.swift
```diff
@@ -268,7 +268,7 @@ public struct HoneycombOptions {
         private var unhandledExceptionInstrumentationEnabled: Bool = true
         private var networkStatusTrackingEnabled: Bool = true
         private var otelMetricKitInstrumentationEnabled: Bool = false
-        private var otelUrlSessionInstrumentationEnabled: Bool = true
+        private var otelUrlSessionInstrumentationEnabled: Bool = false
```
MustafaHaddara, line 271: this is documented in the README and we should update that

PR #136, opened 2026-09-02
fix: URLSession delegate proxy crash on iOS 27

benhanzl:

Fixes an issue where any screen using SwiftUI's `AsyncImage` terminates the app on iOS 27.

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException',
reason: '-[_TtC7SwiftUI...AsyncImageDownloader URLSession:task:didCompleteWithError:]:
unrecognized selector'
```

`ProxyURLSessionTaskDelegate.responds(to:)` returned `true` for `urlSession(_:task:didCompleteWithError:)` but the class never implemented it, and `forwardingTarget(for:)` handed every unhandled selector to the wrapped delegate without checking whether it responded. When the wrapped delegate also lacked the optional method, the message reached the end of the forwarding chain and trapped. Note the class named in the crash is the *wrapped* delegate, which is why this reads at first like a SwiftUI bug.

### Why it stayed latent until now

URLSession only sends `didCompleteWithError:` to a delegate for **delegate-driven** tasks. Tasks with a completion handler (and `async`/`await`) receive completion via the block and never send it.

iOS 27 reworked `AsyncImage` to support HTTP caching, `URLRequest` init, and a caller-supplied session via `asyncImageURLSession(_:)`. That requires a delegate-driven data task, so SwiftUI's `AsyncImageDownloader` now attaches a per-task delegate implementing the data-delegate methods it needs but not `didCompleteWithError:`. `URLSessionTask._instrumented_resume()` only wraps when `self.delegate != nil` — previously `nil` for AsyncImage, now not.

The blast radius is wider than AsyncImage: it also hits `URLSession(configuration:delegate: nil, delegateQueue: nil)` (the proxy wraps `nil`, so `forwardingTarget` returns `nil`) and any third-party delegate without the method, on **all** iOS versions.

- **`responds(to:)` no longer hardcodes selectors.** It now derives the answer from `super.responds(to:)` (which covers whatever this class actually implements) plus the wrapped delegate, so it can't drift out of sync with the methods below it again.
- **`forwardingTarget(for:)` only returns a delegate that responds**, instead of forwarding blindly into a trap.
- **Implemented `urlSession(_:task:didCompleteWithError:)`**, forwarding to the wrapped delegate via the optional-method call.

Implementing that method also ends the span for tasks that **never collect metrics**, which previously leaked, since spans were ended only in `didFinishCollecting`. Failed and cancelled tasks were already covered. Because both callbacks can now fire for the same task, the span is taken from the task and cleared, so it ends **exactly once**.

### Record transport errors

Review of the above turned up an adjacent gap. `updateSpan` only ever set `http.response.status_code`, so a request that fails *below* the HTTP layer — a timeout, a refused connection, a DNS failure — has no status code, never reached `updateSpan`, and ended with `Status.unset` and no error attribute.

### Cancellation recorded as error

- **Cancellation is no longer recorded as an error.** `task.cancel()` and `invalidateAndCancel()` deliver `NSURLErrorCancelled`, which the second commit recorded as a span error. `AsyncImage` in a scrolling list cancels image loads constantly — the exact workload this PR targets — as do view teardown and search-as-you-type, so on upgrade users would have seen a step change in client-span error rate representing no real failure. `endSpan` now skips `-999`; the span still ends, just without an error status.

## CI changes

Three commits on this branch are CI-only, prompted by the fact that the crash is iOS-27-specific and nothing in CI ran on iOS 26 or 27.

- **Unit and smoke legs now run on iOS 18.6, iOS 26.5, and the iOS 27.0 preview.** Each leg is a matched (Xcode image, runtime, device) triple rather than a matrix, because CircleCI images ship only the default runtimes for a release plus the previous major, and device types are runtime-specific: `iPhone 17` does not exist under iOS 18.6, and `iPhone 16` is absent from the 27.0.0 image entirely.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,10 +9,24 @@ orbs:
   macos: circleci/macos@2
   bats: circleci/bats@1.0.0
 
+# Xcode image / iOS runtime pairings, for reference when editing the legs below.
+# CircleCI images ship only the default runtimes for a release plus the previous
+# major, and device types are runtime-specific:
+#
+#   xcode: 26.6    -> iOS 18.6, iOS 26.5   (iPhone 16 and iPhone 17 families)
+#   xcode: 27.0.0  -> iOS 26.5, iOS 27.0   (iPhone 17 family only)
+#
+# So `iPhone 17` does not exist under iOS 18.6, and `iPhone 16` does not exist in
+# the 27.0.0 image at all. Each test leg is a matched (image, iOS, device) triple,
+# which is why the legs are listed out individually rather than built as a matrix.
+
 jobs:
   lint:
+    # Lint stays on the GA Xcode image only. `make lint` runs
+    # `swift format lint --strict`, and a beta toolchain's formatter can disagree
+    # for reasons that have nothing to do with the code under review.
     macos:
-      xcode: 26.1.0
+      xcode: 26.6
     resource_class: m4pro.medium
 
     steps:
@@ -23,8 +37,10 @@ jobs:
     parameters:
       destination:
         type: string
+      xcode_version:
+        type: string
     macos:
-      xcode: 26.1.0
+      xcode: << parameters.xcode_version >>
     resource_class: m4pro.medium
     steps:
       - checkout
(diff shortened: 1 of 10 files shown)


MustafaHaddara (approved):
thank you for figuring this out!

PR #141, opened 2026-09-16
maint: remove dead otel-version test matrix

benhanzl:

We are pinning to exact versions of the OTel SDKs and no longer need to run against older versions in CircleCI. We've pinned to exact versions since #108, and the test matrix has been silently failing. This removes the test matrix and cleans up any related branch naming.

- Delete `set-otel-version.sh`.
- Drop the unused `otel_version` parameter from the `ios_smoke_tests` job and the run step that invoked the script.
- Drop the six `otel_version: "2.0.2"` arguments.
- Rename the aggregate check `smoke_tests-2.0.2` → `smoke_tests`, describing what those legs actually verify.

No consumer-visible change; nothing outside CI referenced the script.

- `circleci config validate` passes.
- `circleci config process` shows the three `smoke-tests-ios-*` legs intact, the aggregate as `smoke_tests`, and no `otel_version` anywhere.
- No residual references to `set-otel-version` or `otel_version` remain in the repo.


tanya-coding (approved):

benhanzl:
## Correcting my earlier note: `smoke_tests-2.0.2` **is** a required status check

The PR description says branch protection "suggests nothing is configured." That was wrong — it was an artifact of a non-admin token getting a 404 on the protection endpoint. Protection is active, and this PR is now blocked by exactly the rename risk it flagged.

Evidence, from a controlled comparison:

| PR | Status contexts | Smoke aggregate | `mergeable_state` |
| --- | --- | --- | --- |
| #139 (merged) @ `76db5c5` | 17, all success | `ci/circleci: smoke_tests-2.0.2` ✅ | `clean` |
| #141 (this) @ `ff18b74` | 17, all success | `ci/circleci: smoke_tests` ✅ | **`blocked`** |

Same context count, every check green, `reviewDecision=APPROVED`, `mergeable=true`, `rebaseable=false` (i.e. not behind — it's rebased onto `6558e47`). The only difference between the two is the aggregate check's name.

### What needs to happen

Someone with admin should **replace** `smoke_tests-2.0.2` with `smoke_tests` in `main`'s required status checks, then merge this PR.

⚠️ **Do not list both names at once.** Any given PR produces only one of them, so requiring both would leave every PR waiting on a check that never reports — blocking all merges rather than easing the transition.

The swap is low-risk right now because this is the only open PR in the repo, so nothing else needs rebasing afterwards. That won't stay true, so it's worth doing while the window is clear.

